PR #2, opened 2025-01-08
Initial image layout documentation and fixups

learmj:
Add documentation to the top level image subdirectory which lists all in-tree layouts as well as common informative points.

Each image layout must contain documentation which describes it, along with clear information about partitions, filesystems and configuration options available to the user.

Since we use mke2fs from the host when creating extN filesystems, we should ensure consistency of creation which means telling genimage to use a mke2fs.conf of our choice rather than /etc/mke2fs.conf from the host.

mbr/simple_dual
 Add documentation
 Set default partition sizes
 Clean up scripts
 Add mke2fs.conf

Update examples which use mbr/simple_dual to set the partition sizes they require.

### DIFF
--- a/config/generic64-apt-simple.cfg
+++ b/config/generic64-apt-simple.cfg
@@ -3,6 +3,8 @@
 
 [image]
 layout=mbr/simple_dual
+boot_part_size=200%
+root_part_size=300%
 name=deb12-arm64-min
 
 [system]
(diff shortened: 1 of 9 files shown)


learmj:
Hi. For the args to mkfs.vfat I copied what pi-gen currently uses. I know our make-boot-image uses -s1 but I believe this is for size reduction so not sure if it's required for an 'RPi OS style' layout.

tdewey-rpi (approved):

PR #3, opened 2025-01-17
Install common initramfs hooks

learmj:
By installing hooks prior to generation of the initramfs, it provides rpi-image-gen the ability to customise initramfs contents if required.

Add a new hook to install rpi udev rules in the initramfs. All rules matching the naming scheme will be installed.

### DIFF
--- a/scripts/bdebstrap/customize10-initramfs
+++ b/scripts/bdebstrap/customize10-initramfs
@@ -2,6 +2,11 @@
 
 set -eu
 
+# Install common hooks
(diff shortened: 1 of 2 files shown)


learmj:
Actually, this isn't required. Custom udev hooks seem to be copied into the initramfs automatically.
See ```/usr/share/initramfs-tools/hooks/udev```.
The lack of installation of my custom rule can be explained by a bug in the ordering of when scripts are run by ```scripts/runner``` introduced by 5e3e914
This PR can likely be closed.

learmj:
Not required.

PR #9, opened 2025-01-29
Add support for image and board initramfs hooks

learmj:
It's conceivable that a board or image may require particular functionality in the initramfs. Add support for that by installing their hooks so they may be executed when initramfs images are generated.

Correct rootfs-overlay location in doc.


learmj:
Rather than limiting this to only initramfs hooks, extend the customisation support so that any ```initramfs-tools``` functionality can be customised. This is achieved by recursively copying the contents of ```device/initramfs-tools/``` if it exists in image or board directories. Makes it much more flexible.

PR #11, opened 2025-02-05
Add kiosk example

learmj:
This example helps to illustrate:
 * Installing pkgs into the chroot (customize stage)
 * Creating a systemd service to run an app for the kiosk
 * Enabling the service so it runs at boot

The example uses all other defaults (ie board and image layout).


learmj:
@ghollingworth This relates to what we discussed the other day. Please let me know if you think additional information is required.

learmj:
Cage has VT switching disabled by default, but I've enabled it here (```-s```). I'll change that so that it's not possible to VT switch.

learmj:
VT switching is now disabled which is likely preferable for kiosk deployments.

PR #21, opened 2025-02-19
mbr/simple_dual: switch to using filesystem UUIDs

learmj:
Rather than labels, use filesystem UUIDs for mounting and identification in this image layout.
Using UUIDs has a number of advantages including identification being transparent across LVM, LUKS, raw block device, etc.

Add uuid-runtime to dependencies.


learmj:
@tdewey-rpi fyi

PR #22, opened 2025-02-25
AB: add dynamic boot slot logic and support

learmj:
The chosen bootloader device-tree attributes now drive the AB boot mech to support automatic 'slot' based rootfs mounting on systems that boot from SD/eMMC or NVMe.

Add a template based scheme which can be used to auto-generate a slot helper easily usable from image layouts that wish to support AB. The helper can be used in udev rules which are installed in the initramfs to create the device node alias symlinks for the mapped slot partition block devices under /dev/disk/by-slot/

If an image layout is to use the slot helper it must define the slot indices needed by the AB mech in a simple text file containing key=value pairs to identify the partition numbers for the supported slot components. This file dictates the slot map that will be pinned via udev rules in the initramfs and it must match the order of the respective partitions in the genimage template for the given image layout.

Using information directly from the bootloader in conjunction with the defined partition map provides a 'one source of truth' method for discovery of the active and other slots at run-time.

The installed slot helper supports several different command line arguments and can be used for many slot related operations, including generation of a tryboot fragment suitable for autoboot.txt using the current boot slot partition as a reference. One example use of this could be write out autoboot.txt after a successful tryboot in order to commit a new update.

Other changes made to the ab_userdata layout include:
  Rename rootfs partitions to 'system'.
  Partition images for A or B must not be different in any way. This is
   critically important because they must be able to be written to any slot.


learmj:
This brings very usable AB functionality to rpi-image-gen with the slot block devices appearing as:
```
$ ls -l /dev/disk/by-slot/
total 0
drwxr-xr-x 2 root root 80 Feb 25 17:41 active
drwxr-xr-x 2 root root 80 Feb 25 17:41 other
```
```
$ ls -l /dev/disk/by-slot/active/
total 0
lrwxrwxrwx 1 root root 18 Feb 25 17:41 boot -> ../../../mmcblk0p3
lrwxrwxrwx 1 root root 18 Feb 25 17:41 system -> ../../../mmcblk0p5
pilot@frobulator:~$ ls -l /dev/disk/by-slot/other
total 0
lrwxrwxrwx 1 root root 18 Feb 25 17:41 boot -> ../../../mmcblk0p2
lrwxrwxrwx 1 root root 18 Feb 25 17:41 system -> ../../../mmcblk0p4
```
The initramfs indicates the system slot delegation as follows:
```
Success: AB: delegated /dev/disk/by-slot/active/system
```
If autoboot.txt is installed on /bootfs, after a ```sudo reboot '0 tryboot'``` to boot into the other slot, if the new system is deemed valid, ```rpi-slot``` may be used to write out an autoboot.txt fragment for the newly system, therefore being able to 'commit' booting from the current slot as the device default , eg ```rpi-slot -t | sudo tee /bootfs/autoboot.txt```.

learmj:
@tdewey-rpi @timg236 fyi

learmj:
v2 contains some more informative msgs on error in the slot template, plus some small misc changes in the image generation, eg marking non-critical fs mountpoints with nofail, pre-computing the slot fs UUIDs and using a known mke2fs.conf for the ext4 mkfs so we don't rely on the host.
I intend on merging later today.

PR #23, opened 2025-02-26
apt: Purge in the chroot by default

learmj:
Removing configuration files for packages which are removed in the chroot during installation is likely the best default behaviour.

Example: install dracut after initramfs-tools.

Without purge, the initramfs-tools config dir hierarchy is retained but update-initramfs not present. This causes rpi-image-gen's core initramfs builder hook to fail because it uses the presence of /etc/initramfs-tools/update-initramfs.conf in the chroot to decide whether to invoke update-initramfs or not.

With purge, the initramfs-tools config dir hierarchy is not retained so the hook doesn't run.

The latter is the expected behaviour because the package is not installed.


learmj:
Renamed variable before merging.

PR #31, opened 2025-03-11
Add support for configurable target type

learmj:
In some cases it may be desirable to create a tarball of the device filesystem rather than a directory (eg for Docker image development). To support this, add a new sys build variable which can be used to specify a path to the target object that bdebstrap will create to hold the device filesystem. The value of this variable is passed directly to bdebstrap via --target and can be a path to a directory (default) or a file. If a file, rpi-image-gen will terminate after bdebstrap execution completes as later stages (eg hooks and genimage) only support operating on a directory.

To support this, rather than pinning the bdebstrap hooks to operate on a directory, pass the chroot directory from bdebstrap to the hooks instead allowing them to operate on the chroot regardless of whether the target is a file or a directory (if a file, a temporary directory will be used and deleted upon bdebtrap termination).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -359,28 +359,31 @@ fi
    --name "$IGconf_image_name" \
    --hostname "$IGconf_device_hostname" \
    --output "$IGconf_sys_outputdir" \
-   --target "${IGconf_sys_workdir}/rootfs" \
-   --setup-hook 'bin/runner setup "${IGconf_sys_workdir}/rootfs"' \
-   --essential-hook 'bin/runner essential "${IGconf_sys_workdir}/rootfs"' \
-   --customize-hook 'bin/runner customize "${IGconf_sys_workdir}/rootfs"' \
-   --cleanup-hook 'bin/runner cleanup "${IGconf_sys_workdir}/rootfs"'
+   --target "$IGconf_sys_target"  \
+   --setup-hook 'bin/runner setup "$@"' \
+   --essential-hook 'bin/runner essential "$@"' \
+   --customize-hook 'bin/runner customize "$@"' \
+   --cleanup-hook 'bin/runner cleanup "$@"'
+
+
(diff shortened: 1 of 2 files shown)


learmj:
Hi @XECDesign 
This should give you what you need to create tarballs using rpi-image-gen, e.g.
```
$ mkdir -p ~/scratch/work
IGconf_sys_target=~/scratch/work/myrootfs.tgz ./build.sh
```
By default, ```IGconf_sys_target``` is a directory (see https://github.com/learmj/rpi-image-gen/blob/feat_flexi_target/sys-build.defaults#L46). If it's a file like in the example above, after bdebstrap execution is complete, rpi-image-gen will deliberately exit meaning no bootable image will be created. This is because subsequent stages like post-build hooks and genimage can only operate on a directory as their input. One downside with this is that because the SBOM is created in post-build, if ```IGconf_sys_target``` specified a file there would be no SBOM generated.

An alternative scheme to creating a tarball of the device filesystem would be to retain the default setting of ```IGconf_sys_target``` (directory) and hook up to ```post-build.sh```. That way would likely need elevated privs via podman to ensure correct ownership in the tarball, though.

We still don't have any 'fully featured' profiles in the tree yet so we can't support anything like an RPiOS lite image created by rpi-image-gen, but for testing you should be able to add your packages to one of the current layers to get a feel for whether this scheme works for you, eg https://github.com/raspberrypi/rpi-image-gen/blob/master/meta/rpi/misc-utils.yaml

Lastly, please note that fstab is fixed up in genimage execution which doesn't run if  ```IGconf_sys_target``` is a file, so ```/etc/fstab``` will be unconfigured in the tarball. Same goes for ```/boot/firmware/cmdline.txt```. There are ways to workaround this though such as using custom bdebstrap hooks with an external dir with ```-D``` which can populate these files. The readme has more info on this, but give me a should if you need.

I'll probably merge this PR as-is unless there is anything glaringly missing because it adds more flexibility to what can be generated by the tool, but do let me know if it works for you or if you have any issues/concerns.

PR #33, opened 2025-03-13
Align all partitions to 8M

learmj:
Higher-capacity cards (particularly Raspberry Pi 128GB A2) have an 8MB preferred erase size, which is a good indication of the flash block size. Other large cards may also benefit.

Refs:
pi-gen pull/833


learmj:
fyi @P33M 
https://github.com/RPi-Distro/pi-gen/pull/833

P33M:
I take it this does the equivalent thing for all specified partitions?

learmj:
Yep: https://github.com/pengutronix/genimage/#hdimage

P33M (approved):

PR #34, opened 2025-03-14
AB(emmc): write-protect the partition table

learmj:
Write-protecting the partition table adds extra robustness. Partitions are currently aligned to 8M so pin the GPT entries to start at +8M, calculate the number of blocks needed to protect this region and lock them using Power on Write Protect via initramfs.

### DIFF
--- a/image/gpt/ab_userdata/bdebstrap/customize03-emmc-ops
+++ b/image/gpt/ab_userdata/bdebstrap/customize03-emmc-ops
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+[[ ${IGconf_device_storage_type:?} != emmc ]] && exit 0
+igconf isy image_ptable_protect || exit 0
+
+chroot $1 apt install mmc-utils
+
+# meta/ab-initramfs has already installed initramfs-tools
+cat <<- 'EOF' > $1/etc/initramfs-tools/hooks/ab-install-files
+#!/bin/sh
+case $1 in
+   prereqs) echo ""; exit 0;;
+esac
+. /usr/share/initramfs-tools/hook-functions
+copy_exec /usr/bin/mmc
+EOF
+chmod +x $1/etc/initramfs-tools/hooks/ab-install-files
+
+install -m 0755 -D ../device/emmc-wp.sh $1/etc/initramfs-tools/scripts/init-bottom/emmc-wp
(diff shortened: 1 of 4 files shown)


learmj:
Hi @P33M @timg236 

What do you think of this? The GPT entries now start at offset +8M so we calculate the number of blocks needed to write protect everything before this (ie the partition table). I think my sums are right. Do you agree?

One downside of doing this is that if a device has a larger WP group (eg 16M instead of 8M), then the script will protect 16M instead of 8 which, based on the genimage layout, will protect part of the first partition.

I do think we should protect the partition table from a robustness point of view, but doing it puts us somewhat at the mercy of device geometry. We can't know that ahead of time when creating the image, so there is an assumption.

The shell-ness needs to work in a posix env and seems fine in klibc (used in initramfs). The slot check just makes sure we've booted from emmc (which of course will apply for sd, too) but the script is only installed for a build that targets emmc storage devices (```IGconf_device_storage_type``` == emmc).

On CM5:
```
Begin: Running /scripts/init-bottom ... Success: WP 1 16 8192 1 16384
```
```
$ sudo mmc writeprotect user get /dev/mmcblk0
Write Protect Group size in blocks/bytes: 16384/8388608
Write Protect Groups 0-0 (Blocks 0-16383), Power-on Write Protection
Write Protect Groups 1-931 (Blocks 16384-15269887), No Write Protection
```

learmj:
> One downside of doing this is that if a device has a larger WP group (eg 16M instead of 8M), then the script will protect 16M instead of 8 which, based on the genimage layout, will protect part of the first partition.

This can be prevented by inspecting ```WP_NUM_BLK``` at the end of the calculation and only issuing the protect op if the number of blocks is <= 8M. Unlike the device geometry/layout, a block is always 512 bytes so if this number is greater than n * 512, where n is 16384, issuing the protect op is skipped. I'd expect a smaller WP group size (eg < 8M) to allow protection of a 8M region because it would be a multiple of 8. My concern is only with devices that may have a WP group size > 8M, so if the check was added it would be at least prevent the first partition being partially locked. I'll make that change and re-submit.

learmj:
v2 contains an overrun safeguard.
Verified the protection scheme functions correctly on both cm4 and cm5 (AGUF4R 14.6 GiB and 8GTF4R 7.28 GiB respectively).
I'll merge this unless there any objections/concerns.

PR #39, opened 2025-03-28
Initial provisioning support

learmj:
This PR adds support for more attributes to the JSON image description, including partition table information and a _provisioning map_. The provisioning map is intended to serve as an input to rpi-sb-provisioner to describe how a device should have the partition table and respective image assets written to it for a given configuration. The provisioning maps are currently standalone JSON files which exist under the appropriate image layout directory and contain the initial version of the JSON describing how a provisioning system can lookup and utilise information from the JSON ```partitionimages[]``` and ```partitiontable[]``` arrays to provision a device.

There are two types of map proposed:

```slotted``` describes an 'AB' system
```flat``` describes a simple RPiOS style system

There are also two flavours of provision map for each in-tree layout:

```provisionmap.json``` describes the layout without any encrypted filesystems
```provisionmap-crypt.json``` describes the layout with encrypted filesystems for root

In an AB system, the provisioning map is used as an input for the udev slot helper to create the pre-defined static mapping of block device nodes to slot sub-components under ```/dev/disk/by-slot/```. This has been extended to include Device Mapper (dm) udev events which are required if a slot has an encrypted sub-component (eg system / rootfs).

### DIFF
--- a/bin/image2json
+++ b/bin/image2json
@@ -6,28 +6,29 @@ import argparse
 import os
 import copy
 import sys
+import subprocess
 
 
-VERSION = "1.4.4"
+VERSION = "2.0.0"
 
 
 IMAGE_KEYS = {"IGconf_device_class",
               "IGconf_device_variant",
               "IGconf_device_storage_type",
-              "IGconf_image_name",
               "IGconf_image_version",
               "IGconf_sys_outputdir"}
 
 
 top_template = {
     "IGversion": VERSION,
-    "meta": [],
+    "IGmeta": [],
     "attributes": {
+        "image-name": "unknown",
         "image-size": "0",
     },
     "layout": {
-        "partition-table-type": "none",
-        "partitions": []
+        "partitiontable": [],
+        "partitionimages": []
     }
 }
 
@@ -180,13 +181,14 @@ def find_images(itype, data):
     return timages
 
 
-# Returns the size of a file. If file is not an absolute path, use
+# Returns the path of a build asset. If the asset is not an absolute path, use
 # the env var as the directory of where to look.
-def get_artefact_size(file, dir_evar):
+def get_artefact_path(file, dir_evar):
     if not os.path.isabs(file):
         dir = os.getenv(dir_evar)
         if not dir:
             raise ValueError(f"Not an env var: {dir_evar}")
+            return None
 
         if not os.path.isdir(dir):
             raise FileNotFoundError(f"Dir not found: {dir}")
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

**Comment by learmj** [comment]
Hi @tdewey-rpi @roliver-rpi @timg236 

Quite a lot here but the main thing I'd like your input on is the proposed JSON ```provisionmap*.json``` files to describe clear and encrypted configurations. Let me know if you'd like any changes, else we can start using these on the provisioner side. I would prefer to create a defined schema for ``flat``` and ```slotted``` provision map system types.

image/mbr/simple_dual/device/provisionmap.json
image/mbr/simple_dual/device/provisionmap-crypt.json
image/gpt/ab_userdata/device/provisionmap.json
image/gpt/ab_userdata/device/provisionmap-crypt.json

Thanks,
--  Matt

File: bin/image2json [hunk 1]
```diff
@@ -60,13 +62,13 @@ def map_genimage(key, value):
     gmap = {
         # https://github.com/pengutronix/genimage#the-image-section
         "partition-type-uuid":{
```
roliver-rpi, line 64: Is CA7D7CCB-63ED-4C53-861C-1742536059CC required here for LUKS?

learmj: For translation support , yes - you're probably right. However, genimage (even looking at current tip) appears to have no concept of LUKS and doesn't support creation of a LUKS container in the image. My working premise with ```image2json``` is that if genimage can parse a config given to it, this script should be able to parse the config, too. The py map referenced will translate a genimage GPT partition type _alias_ into a  'real' GPT partition type UUID, but as there is no alias for a LUKS partition there is no entry for that type in the mapping table.

If genimage was describing a LUKS container (ie a partition) with logical volumes inside, then it should definitely have the ability to describe the container with the correct UUID, but it doesn't have that support. A PR into the genimage maintainers to add a ```luks``` alias would likely need to be submitted alongside LUKS image support which is itself a much larger task.

Having that support would be great because we would have the ability to create encrypted disk images offline rather than relying on the provisioning running on-device to create them. I can see this as being a potential issue for working with generating images on a previously provisioned device, ie the only way to write an image to the device would be to do it via the provisioner. This may be sub-optimal in some cases.

With things as they stand, I think the correct thing to do for ```rpi-image-gen``` is to probably identify (e.g. using a provisioning map such as ```provisionmap-crypt.json```) the partitions which are to exist inside a LUKS container as type ```lvm``` (```e6d6d379-f507-44c2-a23c-238f2a3df928```). That way, when those partition images gets written to the device by the provisioner, they have the correct attribute (since they will be sub-volumes inside the LUKS container).

learmj:
> A PR into the genimage maintainers to add a luks alias would likely need to be submitted alongside LUKS image support which is itself a much larger task.
...
> Having that support would be great because we would have the ability to create encrypted disk images offline

fwiw I think it would be possible for a genimage config to make reference to a 'luks image' and for this to be be generated manually, therefore allowing it to be referenced by a genimage config. If a partition makes reference to an 'image' filename, genimage generates an implicit file rule for it which means it must exist, but genimage doesn't have to create it.

PR #48, opened 2025-04-24
SBOM: Update syft to 1.22.0 and add yaml config

learmj:
Add a file to hold the settings syft will use to generate the SBOM. The path to this file is held in a new config variable.

Remove the config variable holding the output format that the SBOM was generated in. This is no longer needed since the output format is now defined in the config file.

A config file is always provided to syft for its scan.


learmj:
@tdewey-rpi fyi

Issue #50, opened 2025-04-30
Invalid image sizes when using dynamic partition size


martin-ottens:
I have built an image with the `mbr/simple_dual` config, ext4 root file system, and set `root_part_size` to 120%.
The output image is completely OK, but there are cases where the size of the image is not a multiple of 512 bytes, so that the Raspberry Pi Imager, for example, reports an error.

A short example:
```
$ wc -c < output.img
1504609894

$ fdisk -l output.img
Disk output.img: 1.4 GiB, 1504609792 bytes, 2938691 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x00000000

Device      Boot  Start     End Sectors  Size Id Type
output.img1 *     16384  114687   98304   48M  c W95 FAT32 (LBA)
output.img2      114688 2949119 2834432  1.4G 83 Linux
```
The size of the image file is `1504609894` bytes, but only `1504609792` bytes are used, 102 bytes are unused and have zero values (also `1504609894 % 512 = 102`).
I suspect that this is actually a bug in genimage, but I'm surprised that the problem doesn't seem to have appeared there yet.

After the image creation is completed, I've implemented the following workaround in my workflow:
```
truncate -s %512 <IMAGE.IMG>
```
This adds zeros to the end of the image file so that a multiple of 512 bytes is reached. This could also be integrated into the post-image hooks if there is no other/better solution.

learmj:
Hi @martin-ottens Thanks very much for the report. I occasionally use rpi-imager to write an image, but these days I typically use [pi-gen-micro's](https://github.com/raspberrypi/pi-gen-micro) fastboot gadget and use a fastboot client to write the sparse image that rpi-image-gen creates to the device. I've not seen this problem before when I've Imager. The genimage ```%``` sizing is a useful feature, but it's very approximate and I can well believe it could lead to what you describe...

> Disk output.img: 1.4 GiB, 1504609792 bytes, 2938691 sectors
..
> The size of the image file is 1504609894 bytes, but only 1504609792 bytes are used
 
[...]

PR #51, opened 2025-05-02
Correct output image file size

martin-ottens:
Ensure for rpi-imager that all generated images have a size that is a multiple of 512 bytes.

See #50

### DIFF
--- a/device/build.defaults
+++ b/device/build.defaults
@@ -38,3 +38,7 @@ ssh_user1=n
 
 # The storage type used by the device. Supported values are sd, emmc or nvme.
 storage_type=sd
+
+# Logical sector size for the uncompressed full image. The output fill size
+# will be a multiple of this value to ensure compatability with other tools
+sector_size=512
--- a/device/build.defaults
+++ b/device/build.defaults
@@ -39,6 +39,6 @@ ssh_user1=n
 # The storage type used by the device. Supported values are sd, emmc or nvme.
 storage_type=sd
 
-# Logical sector size for the uncompressed full image. The output fill size
-# will be a multiple of this value to ensure compatability with other tools
+# Logical sector size of the device storage for which the image is intended. 
+# The output raw image file size will be a multiple of this value.
 sector_size=512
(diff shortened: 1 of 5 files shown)


learmj:
Hi. Thanks for your submission. I would prefer to have the truncate value held in a config variable relating to device sector size. Doing so would mean it can be used by rpi-image-gen elsewhere as needed (e.g IGconf_device_sector_size and defined in device.defaults). While it definitely makes sense to ensure the image generated is compatible with the device sector size, I don’t think mentioning rpi-imager in the comment is necessary. Any tool that performs simple sector level IO ops would require the change to help ensure the image is written in entirety.

martin-ottens:
Thanks for the feedback. I have made the sector size configurable.

The configured sector size is now also used for the two genimage configs in `mbr/simple_dual`, so that the magic value there is no longer present.

device/build.defaults
```diff
@@ -38,3 +38,7 @@ ssh_user1=n
 
 # The storage type used by the device. Supported values are sd, emmc or nvme.
 storage_type=sd
+
+# Logical sector size for the uncompressed full image. The output fill size
```
learmj: Pedantic comment: Since this is in the device defaults, IMO this particular help/documentation comment should reflect that this attribute is a device level setting rather than an image level setting (otherwise it would have been in ```image/build.defaults```). I recognise there is crossover between the two here.
What about:
> Logical sector size of the device storage for which the image is intended. The output raw image file size will be a multiple of this value.

learmj:
Hi @martin-ottens 
Thanks for taking the time to re-work things and for reflecting the setting elsewhere. Nice one.
If the comment can be fixed up, I'll happily merge it.
--  Matt

martin-ottens:
I have adapted your suggestion for the comment and of course appreciate the effort to keep comments and the project clean

learmj:
Thanks @martin-ottens  I've merged your changes. In future, unless there is a specific need to do otherwise, a squashed flow would be preferred (ie one commit to address one issue - just make the changes on your branch with ```git commit --amend``` and force-push to your branch).
I appreciate you taking the time to address the issue you reported here and working on a solution.

PR #68, opened 2025-06-25
Improve UUID integration

learmj:
Use our pre-defined UUIDs for as many things as possible. UUIDs enable transparent handling and identification of block device and are agnostic to the underlying media type, eg partitions, LVM, LUKS etc. Using UUIDs across the board where possible is a solid improvement.

Add a new specific object to the PMAP JSON to describe static information of a partition. This is not currently used by IDP and only serves to describe the 'pinned' block devices used for the AB slots.

Add plumbing to the PMAP helper to retrieve the static info and introduce some validation.

Since UUIDs are now embedded into the PMAP, AB systems require them to be created in time to be plugged into the slot helper. To satisfy that, move their generation from image creation to post-chroot as the slot helper is needed in the initramfs.

Migrate the slot helper to use UUIDs rather than partition numbers.

Update the AB image slot udev rules to account for invoking the slot helper with device UUIDs. Remove other rules since they are no longer required.

Linux/udev treats labels specified via mkfs.vfat as a 'UUID' even though they're not 128bit. Simplify things and reduce the hackery needed to translate a UUID to a mkfs.vfat label by providing a dedicated label to the genimage templates.

For crypt PMAPs, specify a UUID for LUKS containers. Provisioning will ensure this is accounted for when the containers are created. Being able to identify a LUKS container by UUID will greatly simplify cryptroot integration.

Remove the dual-crypt AB PMAP. While more than one LUKS container can be supported at provisioning, it's unclear if this is something that will actually be used.

Bump minor versions for PMAPs that have been updated.


learmj:
fyi @tdewey-rpi @roliver-rpi

Issue #79, opened 2025-08-13
Updates and breakages Q3'25


learmj:
rpi-image-gen continues to evolve and while the tool is still very much in its infancy, it has become clear that the philosophy behind its use needs focus and refinement. Several users have expressed confusion and frustration about how to customise their build using the config variables, and functionality one would reasonably expect from a 'build system' is missing (e.g. clean). To try and address these concerns I'm making the following major changes to rpi-image-gen. Some of them do break compatibility with how the tool is currently used, and I apologise for that. I'd really rather avoid causing inconvenience to users of a tool I'd like them to stick with. I believe these changes are for the greater good and will improve the experience for all, as well as making the tool much more powerful, easier to use and smoother for new users to get onboard with.

**What's changing?**

1. Support of the options file has been dropped. The options file was badly documented, subject to zero checking and was effectively a side-channel hack to inject variables into the build. It hasn't been clear to all users what it's for, or how to use it. Lack of variable checking costs time and causes frustration (eg #57). There is no structure to what's in the options file other than ```<known prefix>_<variable>```.

Variables will only be able to be supplied:

a) In the config file
b) On the command line
A combination of both is supported.

Option file migration guide:
If your options file has ```foo_myvar=x``` you can either have...
```
[foo]
myvar=x
```
...or pass ```IGconf_foo_myvar=x``` on the command line.

2. Relocatable execution (advanced preparation for debian packaging). In the future, rpi-image-gen is envisaged to be packaged as a deb. This means it needs to self locate better and be able to resolve paths to internal assets.  

3. rpi-image-gen will be the name of the top level script and not build.sh.
 
4. Custom metadata support and consolidation of 'info files' into layers. This is the big one. rpi-image-gen now leverages a layer approach for everything - device, image, etc. Layers still adhere to the same bdebstrap YAML files but can now describe and declare variables using custom ```X-Env-``` metadata fields based on [DEB822](https://repolib.readthedocs.io/en/latest/deb822-format.html) syntax. Adoption of this makes the tool much more powerful and improves things across the board, allowing support for automatic variable validation, default values, dependencies between layers, inheritance, auto-generation of documentation, etc. New python classes constitute the 'core engine' and manage config file parsing, metadata and layer orchestration.

[...]

PR #85, opened 2025-09-02
Brave new world

learmj:
This PR implements a comprehensive redesign of the tool, addressing all items mentioned in #79. Key changes include:

**New Features:**
- Layer metadata and management via dedicated Python engine
- Layer dependencies and automatic validation
- Variable declaration and metadata inspection
- Defined evaluation policy for configuration settings
- Automatic path-based resolution of layers and config files
- Auto-generated HTML layer reference

**Improvements:**
- Enhanced documentation with comprehensive technical developer/user material
- Robust configuration controls and validation
- Improved configuration processing feedback

This update addresses user feedback, provides a more powerful architecture and tooling solution, and establishes the foundation for the future.


bin/runner
devopsbob, line 1: I like the checks here. Does the +x check on lines 10 and 11 check if the script is executable? Can line 35 say script is not executable rather than skipped? Or does test -x mean other things too? (thinking out loud) curious whether this folder should only contain executable scripts or does it also contain other scripts possibly referenced in the build; would you want to skip files on purpose for testing/validations, or are they skipped because they are sourced from other files (aka readable test?).

device/cm5/device.yaml
devopsbob, line 1: I see the rhythmn for changing cfg to yaml. Much cleaner and easier to read and understand. 

Question here is knowing/finding ways to check packages/firmware you want included, how to find/list those firmwares (yeah, should know this maybe by the packaging/specs of the given machine, but is there a way to list devices before hand via software or is it a hardware only thing? Seems you could brick your machine if you apply the wrong package(s). Feedback only, not something I would expect change in this file (or list revisions/version commented out of common packages for device releases/dates??)

stu-spp (approved):

devopsbob (commented):
Great work here. Oh yeah, the legacy is removed though lots of test yamls present so thank you. Curious that docs are in html rather than markdown but understandable if thats your preferred. Seems minor changes will be harder to capture in html than in md, but that would depend if your using a generator during the build I suppose.

bin/runner
learmj: There are two checks on each bdebstrap script, which will result in execution of the script or the script being:

1. Ignored because its not executable
2. Skipped because its name does not conform

Lines 10 & 11 are different and are testing if those variables are set and if so, `bin/runner` will look for bdebstrap scripts within them.

learmj: The scripts that exist in these directories are general purpose hooks and can do anything, really. They have access to the chroot filesystem being created and run with the same privileges as that of the build process itself. Their main use is injecting or manipulating things in the chroot that might not otherwise be possible (or as easy to do) in layer files. A layer can declare and execute the same 'stage' of script operation (ie essential, customize, cleanup, etc) in its own syntax, ie

layer.yaml
```
mmdebstrap:
  packages:
    - xxx
  customize-hooks:
    - chroot $1 sh -c "<expression>"
```
...is largely identical to:

bdebstrap/customize10-myops
```
#!/bin/sh
chroot $1 sh -c "<expression>"
```

If I had to make a distinction between the two, I'd say that scripts in layers are for simple operations whereas bdebstrap scripts are for more involved operations, or when you need to enforce a particular order of execution in the processing. bdebstrap scripts are executed after layer scripts (for the same stage). You can see the processing order in the final config.yaml that's written by the build engine.

I should probably add that to the docs :-)

device/cm5/device.yaml
learmj: Thanks. Yes, I wanted to migrate everything to one file format. Since the layers are already declared in yaml, it makes sense to use it for everything. It's nice and declarative.

Re: packages - Yes you have to know that the package you install is compatible for your device. Regarding arch specifically, Debian will take care of that for you. Regarding package contents, it's largely up to the maintainer/distributor of those packages. For Raspberry Pi packages (eg [raspi-firmware](https://github.com/raspberrypi/firmware)), the assets installed by that pkg are compatible with any Raspberry Pi device. If there was a package that only installed boot firmware for device A, I'd expect that to be made obvious in the pkg name and in the pkg description.

learmj:
> Great work here. Oh yeah, the legacy is removed though lots of test yamls present so thank you. Curious that docs are in html rather than markdown but understandable if thats your preferred. Seems minor changes will be harder to capture in html than in md, but that would depend if your using a generator during the build I suppose.

Thanks. Docs exist in both asciidoc and html. It's only the layer file 'catalogue' that's html only. The reason for that is because of the navigation links in those docs and a desire to present the information in an organised and navigable way (eg links to dependencies, reverse dependencies, variable policies, as well as having a nicer layout in which to show all the layer info - architectures, packages, variable table, as well as embedded companion doc info, etc).

learmj:
Sneaking in a new feature...

PR #102, opened 2025-09-23
core: Support build without /sbin in $PATH

jspricke:
On Debian /sbin is not in $PATH by default. Add it so the tools are
found.

genimage calls sbin/mkdosfs and other tools in sbin.

### DIFF
--- a/bin/image2json
+++ b/bin/image2json
@@ -322,7 +322,7 @@ def parse_genimage_config(config_path):
     img_path = get_artefact_path(img_name, "IGconf_image_outputdir")
     try:
         res = subprocess.run(
-                ["sfdisk", "--verify", "--json", img_path],
+                ["/usr/sbin/sfdisk", "--verify", "--json", img_path],
                 capture_output=True,
                 text=True,
                 check=True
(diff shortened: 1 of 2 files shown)


learmj:
This is unnecessary if PATH is set up appropriately in the first place. For example, on Ubuntu, my regular user's path already contains `/usr/sbin` meaning the user is able to run mkdosfs, fdisk, etc. Adding specific directories to PATH makes the code less portable. While I appreciate the submission, it's not accepted. 

What may be beneficial is that after installing `<pkg>` to provide `<program>` as part of setting up the dependencies, `<program>` could be tested (eg with `command -v <prog>`) to see if it can be found and if not, an error raised (i.e. implies not found in PATH). However, this assumes that after `<pkg>` installation, `<program>` is be available in PATH (should happen with dpkg, but never used to and instead may have required the user to log out and in again). The safest and most portable (and supportable) solution for rpi-image-gen is to expect that all required programs can be found in PATH for the regular user running the build. As rpi-image-gen does not require sudo to run, this should be straight-forward.
Is adding a tiny fragment in your build host `/etc/profile.d/` or shellrc so bad?

jspricke:
It is actually enough to run
```
PATH=/sbin:$PATH ./rpi-image-gen
```
Do you have a proposal where to document that?

(Closing as not needed.)

PR #104, opened 2025-09-26
Improve 4Kn support

learmj:
Improves handling of the config variable that defines the target device storage sector size (basically now enforces 512 or 4096)
Encourages mkfs.ext4 to use a minimum block size of 4K if that variable indicates a 4Kn device.

### DIFF
--- a/docs/layer/device-base.html
+++ b/docs/layer/device-base.html
@@ -286,7 +286,7 @@
                            
                        
                     </td>
-                    <td>Must be one of: 512, 4096</td>
+                    <td>Integer value in range 512 to 512</td>
                     <td>
                         <a href="variable-validation.html#set-policies" class="badge policy-immediate" title="Click for policy and validation help">immediate</a>
                     </td>
--- a/docs/layer/device-base.html
+++ b/docs/layer/device-base.html
@@ -123,7 +123,7 @@
     <div class="header">
         <h1>device-base</h1>
         <span class="badge">device</span>
-        <span class="badge">v1.2.0</span>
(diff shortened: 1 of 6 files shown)


learmj:
Hi @martamomotko @timg236 @roliver-rpi 

The only functional difference here is if `IGconf_device_sector_size` is 4096, `-b -4096` is added to the mkfs.ext4 args to ensure a minimum block size of 4K for all ext4 filesystems.

`IGconf_device_sector_size` is already being passed to mkfs.vfat via `-S`.

learmj:
Hi @martamomotko @timg236 
From https://github.com/raspberrypi/rpi-eeprom/issues/577, I assume there are no issues if I merge this now?

timg236:
@martamomotko Did you manage to get the rpi-image-gen image to boot on NVMe, it sounded like there was a problem with the MBR  / offset of the FAT

martamomotko:
Hi @learmj 
From what I can tell, the 4Kn image generation seems to have incorrect partition alignment. The GPT header offset is at 0x200(512) instead of 0x1000(4096), which means it assumes 512-byte sectors.

learmj:
Hi @martamomotko 
As discussed, we cannot currently use rpi-image-gen to generate 4Kn compatible device images. Therefore setting `IGconf_device_sector_size` to anything other than 512 must result in a configuration validation error.
The other changes on this branch are valid for setting filesystem attributes.
The genimage templates can be revisited once 4Kn support is implemented in upstream genimage.

martamomotko:
Yes, thank you @learmj

PR #111, opened 2025-10-10
Add example splash screen support

tdewey-rpi:
- Introduced a new directory for splash screen examples, including a configuration file and documentation.
- Added a YAML configuration file for splash screen settings, allowing customization of the splash image and boot parameters.
- Created documentation detailing the features, usage, and technical requirements of the splash screen layer.
- Included an example script for generating a simple splash screen image using ImageMagick.


layer/rpi/splash-screen.yaml
```diff
@@ -0,0 +1,81 @@
...
+      
+      # Copy the image into the chroot temporarily
+      TEMP_IMAGE="/tmp/splash-screen-source.tga"
+      cp "$SPLASH_IMAGE" "$1$TEMP_IMAGE"
+      
+      # Run the configure-splash script inside the chroot
+      echo "Configuring splash screen from $SPLASH_IMAGE" >&2
+      
+      # Build command arguments
+      if igconf isy splash_skip_image_checks && ! igconf isy splash_update_cmdline ; then
```
learmj: It's fine to use igconf here to test variables, but it's more defensive to use set -u and reference the variable directly (eg testing `$IGconf_splash_skip_image_checks`). For example, if the X-Env-Var declaration was changed or there was a typo, it would be easy to miss in the scripting logic with igconf, whereas it would result in a hard failure if the variable was referenced by name. Using `X-Env-Var-*-Set = y` will always ensure the variable has a value.

layer/rpi/device/splash-screen/splash-screen.yaml
```diff
@@ -0,0 +1,81 @@
+# METABEGIN
```
learmj, line 1: The splash layer and associated assets would be more applicable under `layer/rpi/device/`, as splash is a device specific feature. I could imagine Connect and OTA etc residing under `layer/rpi/device/services/`.
Just trying to keep compartmentalisation of layers in mind as we add them.

layer/rpi/splash-screen.yaml
```diff
@@ -0,0 +1,81 @@
+# METABEGIN
+# X-Env-Layer-Name: rpi-splash-screen
+# X-Env-Layer-Desc: Raspberry Pi fullscreen splash screen support with custom image configuration.
+# X-Env-Layer-Version: 1.0.0
+# X-Env-Layer-Requires: rpi-essential-base
```
learmj: Does it need rpi-essential-base?

learmj:
Tom - this is fantastic. Thank you. Just one suggestion about the location of the splash layer and assets.

layer/rpi/device/splash-screen/splash-screen.yaml
tdewey-rpi: ACK.

tdewey-rpi: To be clear, however - this is a kernel feature, not a device feature. It's in no way dependent on the Raspberry Pi device you are using.

layer/rpi/splash-screen.yaml
tdewey-rpi: No. Removed.

layer/rpi/device/splash-screen/splash-screen.yaml
learmj: Indeed, but the functionality can only exist on a device which is able to support the feature (needs kernel). The same applies to the ab slot mech layer which can only exist on-device (needs DT info, udev rules, storage, kernel). 

Neither splash nor ab slot mech makes sense to install in a chroot tarball intended to be used as the base for a container (for example) - which would be far less coupled to the underlying device on which it was running on.

rpi-image-gen isn't just for device images, eg it can create a tarball suitable for CI use, or as the base for a cross-build container on an x86 box (see [this layer](https://raspberrypi.github.io/rpi-image-gen/layer/arm64-toolchain-cross.html) which can transparently be pulled in by [this one](https://raspberrypi.github.io/rpi-image-gen/layer/gnu-common.html)).

I'm digressing a bit, and maybe being too pedantic :) 

The above are probably all moot points anyway because layers are located by name rather than by location, so it doesn't actually matter where they reside. The point I'm making relates to directory hierarchy organisation rather than anything else. I suppose if things start to get too cluttered we can reshuffle.

Splash is a great feature to have added, so let's get it in as-is. Thanks again.

learmj (approved):

tdewey-rpi:
Updated to reflect review comments, included an example TGA.

learmj (approved):

learmj:
> Updated to reflect review comments, included an example TGA.

Brilliant. Thanks Tom.

PR #115, opened 2025-10-14
AB: enforce boot‑disk–backed device eligibility

learmj:
Restrict slot candidates to devices backed by the boot disk. This excludes devices on other disks, even if labelled for slots.

Only block devices whose lineage resolves to the boot disk are considered for slot assignment:
- Non‑DM: must be a partition on the boot disk.
- DM: sysfs slaves chain must resolve to a partition on the boot disk.

Devices on other disks are ignored even if they have slot‑compatible GPT labels. Docs updated.

### DIFF
--- a/docs/layer/rpi-ab-slot-mapper.html
+++ b/docs/layer/rpi-ab-slot-mapper.html
@@ -161,6 +161,14 @@
 </div>
 </div>
 <div class="sect1">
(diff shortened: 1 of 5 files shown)


learmj:
Hi @pelwell There should be no functional change incurred as a result of this commit. It just tightens up on the devices we'll accept for slot assignment consideration.

PR #118, opened 2025-10-22
bdebstrap: Install from source to venv

### LINKED ISSUE #113: Building images in Trixie is not possible
As bdebstrap is not available in Trixie by default, there needs to be an alternative method to install it. This should be handled completely transparently by rpi-image-gen and ideally, installed in a location where it can be reused across builds.

Further information can be found here:
https://github.com/bdrung/bdebstrap/issues/16

tdewey-rpi:
- Added a function to install bdebstrap from source into a Python virtual environment.
- Updated the depends file to reflect that bdebstrap is now installed from source.
- Included checks for existing installations and symlink creation for hooks.
- Automatically activates the bdebstrap virtual environment when sourcing dependencies.sh.

Fixes #113

### DIFF
--- a/depends
+++ b/depends
@@ -8,7 +8,7 @@ curl
 genimage
 mtools
 mmdebstrap
-bdebstrap
+# bdebstrap - installed from source in dependencies.sh
 podman
 zstd
 pv
--- a/lib/dependencies.sh
+++ b/lib/dependencies.sh
@@ -1,5 +1,18 @@
 #!/bin/bash
 
+# activate_bdebstrap_venv
+# Adds the bdebstrap venv to PATH if it exists
+activate_bdebstrap_venv()
+{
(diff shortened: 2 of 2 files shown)


tdewey-rpi:
Tested on Trixie, generating images in a manner that should be reliable going forward. Critically, not interfering with the system packages or system installed bdebstrap.

tdewey-rpi:
@learmj I'm not sure if this is the approach you want to take, but I thought I'd offer it up as it gets me out of a bind.

linotypical:
> @learmj I'm not sure if this is the approach you want to take, but I thought I'd offer it up as it gets me out of a bind.

It got me out of a bind. Thanks!

learmj:
Hi @tdewey-rpi Thanks for this. It's not quite the direction I wanted to take. I hope to have an alternative proposal / solution later this week / next.

PR #122, opened 2025-10-29
IDP: Validate provisioning map against schema

learmj:
Introduce schema validation of the Provisioning Map (PMAP). The schema is selected by a new base image layer config variable, with default schemas residing under layer/base/schemas/provisionmap/.

The generic post-image hook now validates the PMAP automatically on every build. This ensures the staged JSON is valid for the provisioning side to process.

Drop (now redundant) checks on some JSON objects in the pmap helper. Schema validation is performed on every invocation.

Clarify some description wording in the AB layer.
Fixup bad partition ref in the AB layer clear pmap for A.system.


learmj:
@tdewey-rpi This schema enables validation of the Provisioning Map component of the Image Description JSON (ie the part which defines how the image should be provisioned on-device).

docs/layer/image-rota.html
```diff
@@ -488,17 +488,17 @@ <h2>Configuration Variables</h2>
                 <tr>
                     <td><code>IGconf_image_pmap</code></td>
                     <td>Provisioning Map type for this image layout.
- All partitions will be provisioned unencrypted (clear).
- System partitions will be provisioned encrypted (crypt).
- System B will be provisioned encrypted (hybrid). Development only.</td>
+ clear: All partitions will be provisioned unencrypted.
+ crypt: All partitions except <slot>:boot will be provisioned encrypted.
+ hybrid: B:system will be provisioned encrypted (development only).</td>
```
tdewey-rpi: We should also consider the reverse-hybrid (read: just-what-matters), where you only encrypt the userdata partition.

learmj: Ack. hybrid is just for testing, but I agree we probably need more granularity regarding what gets encrypted. I'll add some follow-up commits to tweaks the types of pmap that we support here.

learmj:
Hi @tdewey-rpi 
Please see this add-on commit - it adds two new pmaps to cover what you mentioned above. Let me know if you'd like anything else here, or any other changes. Thanks.

PR #133, opened 2025-12-01
Add AB post-image

learmj:
Remove mutual exclusion for post-image hooks - it's not needed.
Plug in a simple post-image hook for the AB layout that packs a compressed archive of slot sparse images.

fyi @pelwell @clowder

### DIFF
--- a/image/gpt/ab_userdata/post-image.sh
+++ b/image/gpt/ab_userdata/post-image.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -eu
+
+upd="${1}/update"
+rm -rf $upd
+mkdir -p $upd
+
+[[ -f ${1}/system.sparse ]] || false
+[[ -f ${1}/boot.sparse ]] || false
+
+ln -sf ../system.sparse ${upd}/system
+ln -sf ../boot.sparse ${upd}/boot
+
+msg "Packing..."
+cd $upd && tar -I zstd -h -cf  ${1}/update.tar.zst -- *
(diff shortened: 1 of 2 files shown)


image/gpt/ab_userdata/post-image.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -eu
+
+upd="${1}/update"
+rm -rf "${1}/update"
```
pelwell: Is there any reason this shouldn't be `rm -rf $upd`?

image/gpt/ab_userdata/post-image.sh
```diff
@@ -0,0 +1,16 @@
...
+mkdir -p $upd
+
+[[ -f ${1}/system.sparse ]] || false
+[[ -f ${1}/boot.sparse ]] || false
+
+ln -sf ../system.sparse ${upd}/system
+ln -sf ../boot.sparse ${upd}/boot
+
+msg "Packing..."
+tar -I zstd -h -cf  ${1}/update.tar.zst  -C $upd .
```
pelwell: This looks neater than the version I was using, but it has the unfortunate side effect of generating entries in the tar file beginning `./`, which causes problems because the name doesn't match that of one of the A/B slots. Could you use this instead?
```
(cd $upd; tar -I zstd -h -cf  ${1}/update.tar.zst *)
```

image/gpt/ab_userdata/post-image.sh
learmj: No. It is now.

image/gpt/ab_userdata/post-image.sh
learmj: Sure.

pelwell (approved):
That works for me!

PR #137, opened 2025-12-03
layer: Fix splash paths and example tga generation

learmj:
- Assets reside under sub-directory device. Update layer and doc. Autogen HTML doc page.
- Example tga generation requires flipping. Because TGAs are are stored bottom‑up by default, imagemagick writes them upside-down (relative to what a viewer expects). Using -flip in the args ensures the splash screen is rendered correctly (in viewer and on-device).
- Update layer category to classify it more appropriately.


learmj:
@tdewey-rpi

tdewey-rpi (approved):

PR #139, opened 2025-12-04
Kiosk example is now AB

learmj:
It makes sense for the webkiosk example to use the AB layout, since this is a reasonable deployment scenario for a remotely managed kiosk.


learmj:
fyi @pelwell @clowder

devopsbob (approved):
Good to support an approach of blue/green deploy or post device deploy access via specific storage mount addressability. Nice simple change for increasing storage for installed packages.

PR #148, opened 2025-12-09
Write artefact version on cleanup

learmj:
/etc/rpi-issue now contains the version identified by config variable IGconf_artefact_version


learmj:
fyi @mudge

PR #149, opened 2025-12-10
Improved startup robustness for the persistent partition

learmj:
Always enforcing repair on the persistent partition filesystem is a best-effort at ensuring it's integrity at startup.
Put `/var/tmp` on tmpfs. This is run-time scratch and does not need to persist to disk.
Mount `/boot/firmware` ro. This filesystem does not need to be writable.
Drop unnecessary journald attribute. No drawbacks. Less noise at boot.

### DIFF
--- a/image/gpt/ab_userdata/image.adoc
+++ b/image/gpt/ab_userdata/image.adoc
@@ -8,14 +8,14 @@ This system uses an A/B, read‑only root with all writable state on a single pe
 |===
 | Mount point | Backing | Type | Notes
 
-| / | /dev/disk/by-slot/active/system | ext4 | Read‑only system root (active slot A or B)
-| /boot/firmware | /dev/disk/by-slot/active/boot | vfat | Boot files (active slot A or B)
-| /bootfs | BOOTFS | vfat | Boot metadata
-| /persistent | PERSISTENT | ext4 | Shared persistent storage
-
-| /var | /persistent/slots/<slot>/var | bind | Per‑slot runtime state (systemd, caches, etc.)
+| / | /dev/disk/by-slot/active/system | ext4 | ro System root (active slot A or B)
+| /boot/firmware | /dev/disk/by-slot/active/boot | vfat | ro Boot files (active slot A or B)
+| /bootfs | BOOTFS | vfat | rw Boot metadata
+| /persistent | PERSISTENT | ext4 | rw Shared persistent storage
 | /home | /persistent/home | bind | User data shared across slots
+| /var | /persistent/slots/<slot>/var | bind | Per‑slot runtime state (systemd, caches, etc.)
 | /var/log/journal | /persistent/log/journal | bind | Single log directory used by both slots
+| /var/tmp | tmpfs | tmpfs | Volatile scratch
 |===
 
 * **Rationale (immutable root + A/B)**
--- a/image/gpt/ab_userdata/image.adoc
+++ b/image/gpt/ab_userdata/image.adoc
@@ -15,7 +15,6 @@ This system uses an A/B, read‑only root with all writable state on a single pe
 | /home | /persistent/home | bind | User data shared across slots
 | /var | /persistent/slots/<slot>/var | bind | Per‑slot runtime state (systemd, caches, etc.)
 | /var/log/journal | /persistent/log/journal | bind | Single log directory used by both slots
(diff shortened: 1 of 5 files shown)


learmj:
fyi @pelwell 
A sanity test would be appreciated. Looks fine on my side.

devopsbob:
/var/tmp is designed to persist between boots differently than /tmp as pure transient. I think tmpfs for /var/tmp will have negative downstream impacts.

image/gpt/ab_userdata/image.adoc
```diff
@@ -8,14 +8,14 @@ This system uses an A/B, read‑only root with all writable state on a single pe
...
-
-| /var | /persistent/slots/<slot>/var | bind | Per‑slot runtime state (systemd, caches, etc.)
+| / | /dev/disk/by-slot/active/system | ext4 | ro System root (active slot A or B)
+| /boot/firmware | /dev/disk/by-slot/active/boot | vfat | ro Boot files (active slot A or B)
+| /bootfs | BOOTFS | vfat | rw Boot metadata
+| /persistent | PERSISTENT | ext4 | rw Shared persistent storage
 | /home | /persistent/home | bind | User data shared across slots
+| /var | /persistent/slots/<slot>/var | bind | Per‑slot runtime state (systemd, caches, etc.)
 | /var/log/journal | /persistent/log/journal | bind | Single log directory used by both slots
+| /var/tmp | tmpfs | tmpfs | Volatile scratch
```
devopsbob: Designed to persist between boots. May have downstream impacts.

image/gpt/ab_userdata/slot-post-process.sh
```diff
@@ -10,14 +10,13 @@ case $COMP in
    SYSTEM)
       cat << EOF > $IMAGEMOUNTPATH/etc/fstab
 /dev/disk/by-slot/active/system /              ext4 ro,relatime,commit=30 0 1
-/dev/disk/by-slot/active/boot   /boot/firmware vfat defaults,rw,noatime,nofail  0 2
+/dev/disk/by-slot/active/boot   /boot/firmware vfat defaults,ro,noatime,nofail  0 2
 LABEL=BOOTFS                    /bootfs        vfat defaults,rw,noatime,errors=panic 0 2
+tmpfs                           /var/tmp       tmpfs mode=1777,nosuid,nodev,size=256M  0  0
```
devopsbob: /var/tmp is designed to persist between boots, putting to tmpfs will have downstream impacts.
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html
https://www.linuxbash.sh/post/mounting-temporary-filesystems-with-tmpfs

image/gpt/ab_userdata/slot-post-process.sh
```diff
@@ -10,14 +10,13 @@ case $COMP in
...
 
 # Bespoke systemd generators mount /persistent, /var and bind mount into it
 # for per-slot storage. See slot-perst-generator.
 
-#tmpfs  /var/tmp  tmpfs  mode=1777,nosuid,nodev,size=256M  0  0
```
devopsbob, line 19: No, not tmpfs for /var/tmp

devopsbob (changes requested):
Please do not use tempfs for /var/tmp.

image/gpt/ab_userdata/slot-post-process.sh
learmj: > I think tmpfs for /var/tmp will have negative downstream impacts.

Such as?

learmj:
Hmm ok
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html

pelwell:
I need /var/tmp to be large enough to hold a full update, even on a 1GB Pi, and it would be nice if we didnt have to re-download it if there's (say) a power cut during the flashing, so here's another vote for /var/tmp to be flash-backed.

learmj:
> I need /var/tmp to be large enough to hold a full update, even on a 1GB Pi, and it would be nice if we didnt have to re-download it if there's (say) a power cut during the flashing, so here's another vote for /var/tmp to be flash-backed.

Ack. I should've checked FHS guidelines beforehand...

devopsbob (approved):

pelwell (approved):

Issue #160, opened 2026-01-15
Disabling journal for system partition on image-rota layer


andrewheberle:
I am working on a process to deploy "delta" updates to "image-rota" devices and have been experimenting with https://github.com/lengfeld/blockdiff for this purpose.

The issue I have come across is that this tool, and other similar ones, cannot patch a previously mounted system partition as simply mounting the filesystem makes changes due to the ext4 journal, despite it being mounted read only, so any binary patch fails to apply cleanly.

Is it possible (or advisable) to disable the journal on the "system" partition in this case? Or possibly make this a configurable option?

Or is this a terrible idea in practice?

learmj:
Hi @andrewheberle  

> Or is this a terrible idea in practice?

Yes. I can think of a few reasons why...

Blockdiff works at the _raw block level_. It isn't ext4 (or any filesystem) aware which means that for a filesystem on the inactive slot, you risk ending up in a 'valid blocks but broken filesystem' situation. In the context of filesystems...

- Layout sensitivity: ext4 frequently relocates extents and metadata. Two filesystems which are the same at the "file/directory contents" level can differ dramatically on disk. A block delta may be large, may fail to apply cleanly, or produce corruption if the base image isn’t _exactly_ what the delta expects.
- Journal/metadata consistency: patching raw blocks may effectively bypass the journal and leave superblocks, group descriptors, bitmaps, and journals inconsistent with the actual data, even if anything is off by just a little.
- Atomicity/rollback: if your device expects the inactive slot to be bootable, a failed blockdiff patch results in a broken slot with no obvious recovery.
- Updates (general): if patching fails, you have no way of reconstructing the base of which to use as a reference for the patch because the data has been left in an unknown state (so cannot be reproduced offline to generate a patch).
- Integrity mechanisms: dm-verity, signatures, or checksums etc over the filesystem image will be invalidated by a block patch unless you also regenerate the hash tree/signature.
- Size/layout changes: block-level deltas break if the partition size, block size, or alignment changes between builds.

A safer scheme would be to either use an immutable filesystem (eg EROFS) to apply diffs to, or write the full ext4 image for the inactive slot atomically.

Using an immutable filesystem like EROFS is safer for incremental/delta updates because:

- Read-only, journal-free: EROFS has no journal or mutable metadata, so you’re patching a static blob rather than a live filesystem that can be left inconsistent mid-patch.
[...]

PR #163, opened 2026-01-19
layer: image-rota: Add option for erofs for the system partition(s)

andrewheberle:
This PR is related to #160 as it adds "erofs" as a filesystem type for the "image-rota" layer.

By setting `IGconf_image_rootfs_type=erofs` the system partition will be created using `mkfs.erofs`.

The changes in this PR "Work's For Me :tm:" however I'm sure there are more elegant ways to achieve what these changes do.

One item I know is less than ideal is that the `image-rota` layer installs the `erofs-utils` package regardless of if erofs is in use or not, but I was unsure how to make this conditional in the layer.


learmj:
Hi @andrewheberle Thanks for proposing this. I will take a look at some point this week. I'm surprised that the way you install the erofs module works, as the pre-image scripts run after chroot creation (which is after initramfs creation).

I just pushed my 'locally hacked but working' branch to [here](https://github.com/learmj/rpi-image-gen/tree/feat/ab_erofs). There's obviously a lot to fix up, but I copy the module into the initramfs differently (same scheme used elsewhere).

Regarding erofs-utils, we have two options:

1. Add it to the depends file so it's a static host dep prerequisite of the tool (same as btrfs-progs).
2. Build it in the mini-sysroot (like genimage and bdebstrap). This would mean we need to build util-linux as a dep which isn't ideal, but it would mean we can use the latest and greatest version of erofs-utils regardless of whether running on Bookworm (which ships a very old 1.5 from June 2022) or Trixie (which ships a much newer version 1.8 from Aug last year. And the tool would only build it if the image was creating a fs with erofs. From what I've read, it was only around ~ 1.6 that it stabilised and it seems like kernel support was ahead of host tool support for some considerable time.

Also, I would like to set the mkfs args specific to the storage media type (`IGconf_device_storage_type` see [device-base](https://raspberrypi.github.io/rpi-image-gen/layer/device-base.html) so we can leverage performance benefits from erofs' pcluster support [1]. So a case statement keying off `IGconf_device_storage_type` setting the appropriate args accordingly.

[1] https://lore.kernel.org/linux-erofs/20210401032954.20555-5-xiang@kernel.org/T/

andrewheberle:
I’m sure this PR needs a lot of polish :)

I was adding the module in the pre-image script but found that didn’t work right so I changed that in `a5502d0` so it’s added to `$1/etc/initrams-tools/modules` as part of the layer `customize-hooks` if `IGconf_image_rootfs_type` is `erofs`.

I’m sure there’s still better ways to do all of this, but for my own proof of concept it seems to work quite well.

andrewheberle:
I've updated my PR to more closely match your branch with the main difference being with my version it still allows ext4 as an option for the image-rota layer by setting `IGconf_image_rootfs_type=ext4`.

learmj:
I saw, thanks. We will need switchable support. There are a wider scope of changes needed if we are to support this properly. I have erofs-utils 1.8.10 built from src using the mini-buildsys in rpi-image-gen as part of the build if the rootfs type is erofs (so we control the version and don't use a potentially old version from the host distro), configured it to support zstd (which is supported by our kernel) and multi-threading (so it creates compressed images quicker - at the expense of some options, eg dupes). This is transparent to the invocation from genimage. I would like to make sure it's all solid, and also run some benchmarks against different storage media types before it goes anywhere near the tree.

I suggest running with your own changes for the time being. Adding erofs isn't a priority for me right now.

andrewheberle:
I’ll just pretend that all makes sense to me, but I’m happy to run with my own branch internally until you are ready to add erofs support to the tree.

andrewheberle:
Closing this PR so I can maintain my own branch for the time being to avoid unnecessary noise here.

learmj:
Hi @andrewheberle That's fine. I've been working on some internal / engine additions, some of which will benefit configurations that wish to use erofs. I'm hoping for these to land in the tree soon. Thank you for taking the time to write up this PR and for your interest in rpi-image-gen.

PR #167, opened 2026-01-28
Introduce hashed password option

### LINKED ISSUE #82: user creds layer to use chpasswd -e (for hashed passwords)
user creds layer:

```yaml
---
name: rpi-user-creds
mmdebstrap:
  packages:
    - sudo
  customize-hooks:
    - chroot $1 sh -c "if ! id -u $IGconf_device_user1 >/dev/null 2>&1; then
        adduser --disabled-password --gecos \"\"  $IGconf_device_user1;
        fi"
    - |-
      if [ -n "$IGconf_device_user1pass" ] ; then
         chroot $1 sh -c "echo ${IGconf_device_user1}:${IGconf_device_user1pass} | chpasswd"
      fi
    - chroot $1 usermod --pass='*' root
    - chroot $1 sh -c "for GRP in input spi i2c gpio; do
         groupadd -f -r \$GRP;
      done"
    - chroot $1 sh -c "for GRP in adm dialout cdrom audio users sudo video games plugdev input spi i2c gpio ; do
         adduser $IGconf_device_user1 \$GRP;
      done"
    - sed "s/^pi /$IGconf_device_user1 /" $RPI_TEMPLATES/sudo/010_pi-nopasswd > $1/etc/sudoers.d/010_pi-nopasswd
    - mkdir -p $1/etc/profile.d
    - |-
      cat <<- 'EOCHROOT' > $1/etc/profile.d/01local.sh
      #!/bin/sh
      if [ "$(id -u)" -ne 0 ]; then
        PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/games:/usr/games
      fi
      EOCHROOT

```
specifically:
```yaml
      if [ -n "$IGconf_device_user1pass" ] ; then
         chroot $1 sh -c "echo ${IGconf_device_user1}:${IGconf_device_user1pass} | chpasswd"
      fi
```
It would be so great if we could use `chpasswd -e` instead of `chpasswd`:
```
  -e, --encrypted               supplied passwords are encrypted
```

so that we could store passwords hashed in version control.

I'm not sure if this is a good idea, but it was something I was considering a while back. Something to think about.

nyukuru:
It may be useful, for security reasons, to allow the password of the initial user to be set with a pre-hashed value. This is a very simple option to add that functionality.

fixes: #82

### DIFF
--- a/layer/base/device-base.yaml
+++ b/layer/base/device-base.yaml
@@ -36,6 +36,7 @@
 # X-Env-Var-user1-Set: y
 #
 # X-Env-Var-user1pass:
+# X-Env-Var-user1pass-Conflicts: user1hashedpass
 # X-Env-Var-user1pass-Desc: Password required to log into the user1 account. If
 #  empty, the account will be locked. Password requirements are as follows.
 #  At least 8 characters.
@@ -46,6 +47,14 @@
 # X-Env-Var-user1pass-Required: n
 # X-Env-Var-user1pass-Valid: regex:^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{8,}$
 # X-Env-Var-user1pass-Set: n
+#
+# X-Env-Var-user1hashedpass:
+# X-Env-Var-user1hashedpass-Conflicts: user1pass
+# X-Env-Var-user1hashedpass-Desc: A pre-hashed value of the password required to log into the user1 account.
+#  to pre-hash a password use the provided genpasswd script in the chroot scope.
+# X-Env-Var-user1hashedpass-Required: n
+# X-Env-Var-user1hashedpass-Valid: string
+# X-Env-Var-user1hashedpass-Set: n
 #
 # X-Env-Var-storage_type: sd
 # X-Env-Var-storage_type-Desc: Device storage media type.
--- a/layer/rpi/user-credentials.yaml
+++ b/layer/rpi/user-credentials.yaml
@@ -9,7 +9,7 @@
 # X-Env-VarRequires: IGconf_device_user1,RPI_TEMPLATES
 # X-Env-VarRequires-Valid: string,string
 #
-# X-Env-VarOptional: IGconf_device_user1pass
+# X-Env-VarOptional: IGconf_device_user1pass,IGconf_device_user1hashedpass
 # METAEND
 ---
 mmdebstrap:
@@ -22,8 +22,10 @@ mmdebstrap:
         adduser --disabled-password --gecos \"\"  $IGconf_device_user1;
         fi"
     - |-
-      if [ -n "$IGconf_device_user1pass" ] ; then
-         chroot $1 sh -c "printf '%s:%s\n' '${IGconf_device_user1}' '${IGconf_device_user1pass}' | chpasswd"
+      if [ -n "$IGconf_device_user1hashedpass" ] ; then
+         printf '%s:%s\n' '${IGconf_device_user1}' '${IGconf_device_user1hashedpass}' | chroot $1 chpasswd -e
+      elif [ -n "$IGconf_device_user1pass" ] ; then
+         printf '%s:%s\n' '${IGconf_device_user1}' '${IGconf_device_user1pass}' | chroot $1 chpasswd
       fi
     - chroot $1 usermod --pass='*' root
     - chroot $1 sh -c "for GRP in input spi i2c gpio; do


learmj:
Hi @nyukuru 
I think if the tool is to offer this option, it should also provide (or at least describe) how to generate the hashed password in a way which is guaranteed to be compatible with the chroot. For example, generating hashes outside the chroot with different library versions can change defaults (eg, yescrypt vs SHA-512).

What I'm referring to is parity between the tooling/algo used to generate the password, and the chroot. In order to guaranteed this, the hash would need to be generated with the same tooling/algo as would be used when executing inside the chroot, for example by running `mkpasswd`.

I think the logic in the PR is fine, but I think it should be accompanied by some documentation about offline hashed password generation.

nyukuru:
I've added a little note to the option description, is there any other place you think this should be documented? I also wondered if the `user1pass` should acknowledge the precedence rules in its description as well or not

stu-spp:
This is great!

But how do we generate a valid hash? Let's all agree on a process, and then stick that in the documentation! Preferably with an explicit example.

Well done!

nyukuru:
> But how do we generate a valid hash? Let's all agree on a process, and then stick that in the documentation! Preferably with an explicit example.

As mentioned in the updated description, it's easiest through `mkpasswd`. Just running `mkpasswd` will give a password prompt, or you can use it like `mkpasswd <password>`. If you feel like any information needs to be added let me know

The debian package of libcrypt (which is depended on by login) supports most common hashing methods (including yescrypt) and should the user build against a thinner libcrypt then I would expect them to know to change the algorithm to conform to the ones supported by the libcrypt.

learmj:
I've got an idea for passwd generation, eg having a util script in `bin/` that could chroot and 'run some stuff'. Makes it a bit more 'seamless' and at least would provide something that could be mentioned in any documentation, eg "run foo to generate the hashed password'. I'll have a crack it later and see if it works. I'll paste it here for eval to see what you both think.

nyukuru:
Part of the reasoning I would like this change is that it would allow sharing the image building pipeline in a public facing repository without exposing the password secret. 

Would the idea be to create a cli option to run a command from within the resulting image; something like:
`rpi-image-gen run --packages mkpasswd mkpasswd`

learmj:
I understand the requirement.

> Would the idea be to create a cli option to run a command from within the resulting image;

You'd run a util offline against the chroot to generate the (hashed) password, but obviously you'd have to provide the (clear) password to the util. However, you'd do that securely against a representative chroot of your build, then use the hashed password in the build config (or via cmdline override), then build / distribute.

I've been working on some more engine changes, one of which is the ability for variables declared in layers to mark that they conflict with another. For example:
```
# X-Env-Var-user1pass:
# X-Env-Var-user1pass-Conflicts: user1hashpass
#
# X-Env-Var-user1hashpass:
# X-Env-Var-user1hashpass-Conflicts: user1pass
```
The engine will never permit both to be set, which would then simplify the handling logic in the layer:
```
Error: [FAIL] Variables 'IGconf_device_user1pass' and 'IGconf_device_user1hashpass' conflict but both are set (layer: device-base)
Error: pipeline failed
```

I've also noticed that `$` is not escaped properly when passed as a string via config file or cmdline :(
When I was testing offline password generation, my hashed password was ` $6$R6URnFO2NsxIYqeC$tsM6Qw60Png93CltU7Om7vZt67oHqlNWwVN/GzG1gZBUtr/3WWuBwDUVV1fvDut.nYNrFSkhh1wj1sCbclrE10`. This was not handled correctly and needs to be fixed.

These two things need to be in the tree before we can add the functionality in this PR.

nyukuru:
I'll look into see what can be done about the parsing/escaping of the hash tonight.

Are said engine changes merged in or would this pr be blocked until that's implemented?

learmj:
No need, but thank you for the offer. I'm already working on something that will prevent the `$` separators from causing chaos :) I plan on incorporating it into a larger change-set that I've been working on (see https://github.com/learmj/rpi-image-gen/tree/engine_dev) hopefully slated for merge next week.

> would this pr be blocked until that's implemented?

Yes. This PR will need changing once the core engine changes land because:

- `user1pass` and `user1hashpass` will be able to be marked as conflicting in the `device-base` layer (ie mutually exclusive).
- That mutual exclusion means `rpi-user-credentials` can just have a simple `if` conditional for handling the hashed password, i.e. the selection / conflict / resolution logic is performed outside of the hooks, therefore keeping them simple (best practice).

Also, I've already got a new util which can generate hashed passwords against a chroot. It seems to work fine but I need to tidy it up and test further. This should probably be merged as part of the PR. We can co-author or I can attach it.

How does that sound?

learmj:
@nyukuru 
With #179 you can now do this...
```
# X-Env-Var-user1pass:
# X-Env-Var-user1pass-Conflicts: user1hashpass
#
# X-Env-Var-user1hashpass:
# X-Env-Var-user1pass-Conflicts: user1pass
```
..and have simple if logic in the layer without precedence, complication, etc.

Please rebase and resubmit this PR so `device-base` uses the above scheme.

Then in `rpi/user-credentials.yaml` we should probably improve the scripts so they do:
```
if [ -n "$IGconf_device_user1pass" ] ; then
   printf '%s:%s\n' "$IGconf_device_user1" "$IGconf_device_user1pass" | chroot $1 chpasswd
fi
if [ -n "$IGconf_device_user1hashpass" ] ; then
   printf '%s:%s\n' "$IGconf_device_user1" "$IGconf_device_user1hashpass" | chroot $1 chpasswd -e
fi
```
This is better than the current invocation because it avoids `sh -c` (which can mishandle quoting/expansion) and doesn’t interpolate the password into a shell command string.

Thanks,
--  Matt

nyukuru:
I've rebased with the changes onto master, I'm not sure if there's conflicts with #179, in which case just let me know I will rebase again.

learmj:
Thank you. Appreciated. #179 will require merging before this can land.
Do you think we should have `IGconf_device_user1hashedpass` or `IGconf_device_user1passhash` ?
Maybe I'm being too pedantic :)

In the variable Description, we can suggest to use genpasswd to produce the hash, but like @stu-spp [wrote,](https://github.com/raspberrypi/rpi-image-gen/pull/178#issuecomment-3872109090) if you already know the expected format, it's fine to use other tools.

stu-spp:
Haha! I vote `IGconf_device_user1hashedpass`. :smile:

layer/base/device-base.yaml
```diff
@@ -47,6 +48,14 @@
 # X-Env-Var-user1pass-Valid: regex:^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{8,}$
 # X-Env-Var-user1pass-Set: n
 #
+# X-Env-Var-user1hashedpass:
+# X-Env-Var-user1hashedpass-Conflicts: user1pass
+# X-Env-Var-user1hashedpass-Desc: A pre-hashed value of the password required to log into the user1 account.
+#  to pre-hash a password use the provided genpasswd script in the chroot scope.
```
learmj, line 54: Suggest:
Pre-hashed password for user account creation. Generate with the in-tree genpasswd utility, or supply a hash created with the same algorithm used by the chroot.

layer/base/device-base.yaml
```diff
@@ -36,6 +36,7 @@
 # X-Env-Var-user1-Set: y
 #
 # X-Env-Var-user1pass:
+# X-Env-Var-user1pass-Conflicts: user1hashedpass
 # X-Env-Var-user1pass-Desc: Password required to log into the user1 account. If
```
learmj, line 40: Password -> Cleartext password

layer/base/device-base.yaml
```diff
@@ -36,6 +36,7 @@
...
 #  empty, the account will be locked. Password requirements are as follows.
```
learmj, line 41: For user1pass let's have:
If both cleartext and hashed password are empty, the account will be locked.

layer/base/device-base.yaml
learmj, line 54: As above, regarding the comment about the account being locked if both variables are empty.

learmj:
Hi @nyukuru  If the additions to the descriptions can be made, I'll merge this after #179 lands.

learmj:
@nyukuru What is the status of this PR?

PR #178, opened 2026-02-06
utils: Add offline password hash utility

learmj:
Add utility to hash passwords from chroot policy (yescrypt/sha‑512) via stdin/prompt.
Required for #82
Required by #167

### DIFF
--- a/bin/genpasswd
+++ b/bin/genpasswd
@@ -1,0 +1,42 @@
+#!/bin/sh
+set -eu
+
+chroot="${1:?Usage: $0 <chroot_path>}"
+
+# Read password (stdin if piped, else prompt)
+if [ -t 0 ]; then
+   printf "Password: " >&2
+   stty -echo
+   IFS= read -r password
+   stty echo
+   printf "\n" >&2
+else
+   IFS= read -r password
+fi
+[ -n "$password" ] || { echo "Empty password" >&2; exit 1; }
+
+method="sha-512"
+if [ -f "$chroot/etc/login.defs" ]; then
+   m=$(awk '/^[[:space:]]*ENCRYPT_METHOD[[:space:]]+/ {print tolower($2)}' "$chroot/etc/login.defs")
+   [ -n "${m:-}" ] && method="$m"
+fi
+if [ "$method" = "" ] && [ -f "$chroot/etc/pam.d/common-password" ] &&
+   grep -q yescrypt "$chroot/etc/pam.d/common-password"; then
+   method="yescrypt"
+fi
+
+case "$method" in
+   sha512|sha-512) method="sha-512" ;;
+   yescrypt) ;;
+   md5) echo "md5 disallowed" >&2; exit 1 ;;
+   *) method="sha-512" ;;
+esac
+
+if [ "$method" = "yescrypt" ]; then
+   command -v mkpasswd >/dev/null 2>&1 || { echo "mkpasswd(whois) required for yescrypt" >&2; exit 1; }
+   mkpasswd --help 2>/dev/null | grep -q -- '--method' || { echo "mkpasswd lacks --method (need whois/libxcrypt)" >&2; exit 1; }
+   mkpasswd --method=yescrypt "$password"
+else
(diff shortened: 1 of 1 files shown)


learmj:
@stu-spp @nyukuru Let me know what you think and if this works for you. As a quick way to test:

1. Use the util to generate a passwd against the chroot.
2. Change `X-Env-Var-user1pass` validator type to string
3. Change chpasswd invocation in `rpi-user-credentials` to use `-e`.
4. Pass the hash via cmdline or config file (see https://github.com/raspberrypi/rpi-image-gen/issues/82#issuecomment-3859630942)
5. Build and test

bin/genpasswd
```diff
@@ -0,0 +1,42 @@
...
+
+method="sha-512"
+if [ -f "$chroot/etc/login.defs" ]; then
+   m=$(awk '/^[[:space:]]*ENCRYPT_METHOD[[:space:]]+/ {print tolower($2)}' "$chroot/etc/login.defs")
+   [ -n "${m:-}" ] && method="$m"
+fi
+if [ "$method" = "" ] && [ -f "$chroot/etc/pam.d/common-password" ] &&
+   grep -q yescrypt "$chroot/etc/pam.d/common-password"; then
+   method="yescrypt"
+fi
```
nyukuru, line 26: Seems like `"$method" = ""` will never be true here

learmj:
I'm satisfied that this implementation fulfils the current requirements of offline passwd generation and intend on merging this week unless there are objections.

Since this util is not part of build-flow, rpi-image-gen doesn't guarantee that its dependencies are installed (whois mkpasswd/libxcrypt for yescrypt, openssl for sha-512). Users will be required to ensure the util can execute on their host.

stu-spp:
Yes! This is great! Users will use the `genpasswd` utility. Or users can also:
```
x@xps:~/workspace/mt/rpi-image-gen$ mkpasswd --method=yescrypt 123456
$y$j9T$4nD4McKclDHBj2c8Ia86p1$Nn/XX2Bi5PCOFd9yf7Hka.jJxv4AzGJqAfAey6MpQI1
x@xps:~/workspace/mt/rpi-image-gen$
```
And then then in a `device-config.yml`:
```
device:
  ...
  user1hashpass: $y$j9T$4nD4McKclDHBj2c8Ia86p1$Nn/XX2Bi5PCOFd9yf7Hka.jJxv4AzGJqAfAey6MpQI1
  ...
```

Woot!

bin/genpasswd
```diff
@@ -0,0 +1,42 @@
...
+   stty -echo
+   IFS= read -r password
+   stty echo
+   printf "\n" >&2
+else
+   IFS= read -r password
+fi
+[ -n "$password" ] || { echo "Empty password" >&2; exit 1; }
+
+method="sha-512"
```
nyukuru, line 18: The manpage in libxcrypt mentions that it would prefer yescrypt and it's derivates to be the default behavior for newly genned passwords

bin/genpasswd
```diff
@@ -0,0 +1,42 @@
...
+   md5) echo "md5 disallowed" >&2; exit 1 ;;
+   *) method="sha-512" ;;
+esac
+
+if [ "$method" = "yescrypt" ]; then
+   command -v mkpasswd >/dev/null 2>&1 || { echo "mkpasswd(whois) required for yescrypt" >&2; exit 1; }
+   mkpasswd --help 2>/dev/null | grep -q -- '--method' || { echo "mkpasswd lacks --method (need whois/libxcrypt)" >&2; exit 1; }
+   mkpasswd --method=yescrypt "$password"
+else
+   command -v openssl >/dev/null 2>&1 || { echo "openssl required for sha-512" >&2; exit 1; }
```
nyukuru, line 40: you can avoid the openssl dependency here by using `mkpasswd --method=sha512crypt`

bin/genpasswd
learmj: `"$method" = ""` is true when method is unset or set to an empty string.
If awk finds no match, m is empty, so method stays empty/unset and the test is true.

PR #183, opened 2026-02-14
Do not overwrite an existing fstasb, instead prepend to the existing fstab

Bernie:
This fix prevents the basic mbr setup from clobbering any fstab modifications in the generated image when it writes the ROOTUUID and BOOTUUID out to the file.

Instead, the script now reads the existing fstab from the image mount path and appends it back afterwards.

The effect is that the contents of the base image fstab persists into the final image, appended after the generated boot and root images.  This allows a generated image to change default parameters of other fstab entries (e.g. changing the size of the tmpfs, etc) without having the image generation process clobber it without warning.

### DIFF
--- a/image/mbr/simple_dual/setup.sh
+++ b/image/mbr/simple_dual/setup.sh
@@ -8,6 +8,7 @@ ROOTUUID="$3"
 
 case $LABEL in
    ROOT)
+      FSTABCONTENTS=$(<"$IMAGEMOUNTPATH/etc/fstab")
       case $IGconf_image_rootfs_type in
          ext4)
             cat << EOF > $IMAGEMOUNTPATH/etc/fstab
@@ -26,6 +27,7 @@ EOF
       cat << EOF >> $IMAGEMOUNTPATH/etc/fstab
 UUID=${BOOTUUID} /boot/firmware  vfat defaults,rw,noatime,errors=remount-ro 0 2
 EOF
+      echo $FSTABCONTENTS >> "$IMAGEMOUNTPATH/etc/fstab"
       ;;
    BOOT)
       sed -i "s|root=\([^ ]*\)|root=UUID=${ROOTUUID}|" $IMAGEMOUNTPATH/cmdline.txt


learmj:
Hi @Bernie 
I think the reasoning behind the PR is valid, but the implementation isn't something I'd be happy merging.

1. `setup.sh` writes a clean, known-good fstab in the output image. I recognise that the current implementation does not allow user customisation (which I believe is the crux of your PR).
2. `setup.sh` runs under `set -eu`. If `/etc/fstab` doesn't exist in the image (which is possible), the read of the file will fail and `set -e` will abort the entire script resulting in a failed build.
3. The echo of `$FSTABCONTENTS` without quotes will undergo word splitting and glob expansion, so any whitespace formatting in the original fstab is destroyed, and filenames or patterns like * in comments would be expanded.
4. It's generally not a good idea to read the entire contents of a file into a variable in a shell script. For small text files (like fstab) it's probably ok (ie no execessive memory usage, no null bytes, can live with trailing new lines being stripped) but if streaming line by line, best to use native tools and/or redirect syntax. Slurping the whole file into a variable and echoing it back unquoted is significantly less safe than other alternatives.

There are better ways of supporting user customisations to fstab that may be in the chroot before the image generation takes place. For example, it would seem a reasonable approach for the heredoc output to _append_ to `$IMAGEMOUNTPATH/etc/fstab` rather than truncating it, ie change `cat << EOF > $IMAGEMOUNTPATH/etc/fstab` to:
`cat << EOF >> $IMAGEMOUNTPATH/etc/fstab`. Whatever was in `/etc/fstab` in the chroot would be retained, with the layer's own entries being added after. The only obvious downside of this is that for a vanilla Debian minbase, this trivial change would result in something like:
```
# UNCONFIGURED FSTAB FOR BASE SYSTEM
UUID=79bd96fd-8d52-418d-83d4-305e37c8d595 /               ext4 rw,relatime,errors=remount-ro,commit=30 0 1
UUID=2417-7B2F /boot/firmware  vfat defaults,rw,noatime,errors=remount-ro 0 2
```
Functionally, that's fine, but it may not be deemed suitable in other cases.

I wouldn't recommend running any generic hooks to simply `rm` the file between mmdebstrap completion and application of fs overlays either, because the file may have been created legitimately (eg, by a layer or hook). It seems my above suggestion is probably the easiest and safest to implement in order to achieve your goal, albeit with potentially ambiguous comments in the file.

PR #195, opened 2026-03-04
image-rota: Generate dm-verity hash assets for root

learmj:
Utilise veritysetup from the host to generate the hash-tree of the EROFS rootfs, saving the root hash to file. This requires upstream changes to genimage v19, so these are applied as patches via the package build framework.

Remove redundant hinting of initramfs fstype - this is already handled by a common hook.


learmj:
fyi @tdewey-rpi

PR #203, opened 2026-03-09
config: add packages hook and document well-known section names

learmj:
Add a bdebstrap customize hook that automatically installs packages declared via a [packages] config section. Values resolving to existing files are installed as local .deb packages. All others are passed to apt. Docs updated.

This largely obsoletes examples/debstore, which required a custom layer to achieve the same result. The example is retained for cases where finer control over the installation process may be needed, and has been updated to use a trixie base config.

### DIFF
--- a/docs/config/index.adoc
+++ b/docs/config/index.adoc
@@ -261,6 +261,68 @@ Therefore, included files are resolved using:
 1. **Parent directory**
 2. **Search path**
 
+== Section Names
+
+Any section name used in a configuration file maps directly to the `IGconf_<section>_*` variable namespace. A layer can declare a section via `X-Env-VarPrefix` in its metadata. This means that the set of meaningful section names in a build is determined by the layers included. Declaring a section allows a layer to expose a group of configuration variables. Some section names carry specific functional meaning because the build system actively responds to them.
+
+=== device
+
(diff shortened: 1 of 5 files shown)


docs/config/index.adoc
```diff
@@ -261,6 +261,47 @@ Therefore, included files are resolved using:
 1. **Parent directory**
 2. **Search path**
 
+== Section Names
+
+Some section names carry conventional meaning within rpi-image-gen. Using them causes the engine or build hooks to treat their variables in a specific way.
```
pelwell: Would "specific" be a better word than "conventional'? I think you mean that there are certain conventions around some names, but conventional makes them  sound ordinary.

pelwell (approved):
I didn't really look at the implementation, but the changes listed here look great. Apart from the question over the use of the word "conventional", I wholeheartedly approve.

docs/config/index.adoc
learmj: Yes, I thought exactly the same. Moreover, if a layer declares `X-Env-VarPrefix`, this maps to a section with the same name. I'll reword it. Thanks Phil.

learmj:
@pelwell Hopefully this reads better (and is clearer)?

pelwell:
Much better. Ship it!

PR #204, opened 2026-03-10
genimage: Add patch to support sparse extents > 4GB

pelwell:
Images with long contiguous regions of either fixed values or no fixed values have been seen to break the Android sparse image generation. Add a simple patch to fix the long-standing To Do item by splitting large extents into multiple chunks.


learmj:
Marvellous. Merged.

Issue #168, opened 2026-01-28
Documentation & UX Request: Guidance for Replicating Official Images and Multi-Arch Builds


martinn-ronin:
**User Perspective**
As a user familiar with Docker-style declarative builds, I find the transition to `rpi-image-gen` quite steep. While the tool is powerful, the current documentation focuses on "minimal base" examples that don't clearly demonstrate how to bridge the gap to a fully-featured or official-style Raspberry Pi OS image.

**The Problem**
1. Replicating Official Builds: There are no clear templates for building the equivalent of an official "Lite" or "Desktop" image. Starting from a blank slate to reach that level of feature-parity is currently a manual "trial and error" process.
2. Multi-Architecture Complexity: Building for armhf (32-bit) vs. arm64 (64-bit) feels complex. Unlike the simple --platform flag in Docker, it is not immediately clear how to handle architecture-specific package names or configurations within the YAML layers.
3. The "Stripping Down" Workflow: Many users prefer to start with a known-good configuration (like an official image) and strip it down. The current documentation doesn't provide a clear path for this "subtractive" approach in a declarative way.

**Requested Improvements**
- Official Baseline Templates: Please provide official YAML configuration templates that exactly replicate the current "Lite" and "Standard" Raspberry Pi OS releases (for both architectures).
- Cross-Arch Guide: A dedicated section or "Cheat Sheet" explaining how to manage multi-arch builds, specifically how the --architecture flag interacts with package selection in YAML.
- Advanced Customization Examples: More complex examples showing how to transition from a minbase build to a production-ready OS with standard utilities (networking, SSH, common drivers) pre-configured.

**Additional Context**
Providing these "Baseline Templates" would allow users to treat them as a "Base Image" (similar to a Docker FROM instruction), making the tool much more accessible for building production-grade customized images.

learmj:
Hi @martinn-ronin Thanks for your questions and interest in rpi-image-gen. My initial response to your problem items are below. I'm more than happy to work things through so you can get to where you need to be with the tool, but first:

> 1. Replicating Official Builds: 

There are intentionally no configurations in the tree at the moment which build an official Raspberry Pi OS image. At the current time, there are no immediate plans to move generation of the official images from [pi-gen](https://github.com/RPi-Distro/pi-gen) to rpi-image-gen.

(I'll answer 3 next)

> 3. The "Stripping Down" Workflow: 
> Many users prefer to start with a known-good configuration (like an official image) and strip it down. 

[...]

PR #207, opened 2026-03-12
doc: add a guide on getting started using configuration files

### LINKED ISSUE #206: Make learning curve more gentle (PR contribution offer)
With rpi-image-gen having reached a nice level of maturity and stability, we are adopting it for creating our Raspberry Pi-based purpose-specific devices.

As part of this process, I would be happy to address some of the challenges we stumbled over, by contributing PR(s) to help future adopters:

1. Add a second level of guided usage
The current documentation targets first a four-line quick start which is great for enticing us, and then essentially throws you in the deep end starting with a section on Philosophy, which is great for developing for the project or using it as a framework. My proposal is to add a simple guide that falls in between, focusing on using rpi-image-gen as a tool rather than a framework. It would be <50 lines and fit in the main README or examples/README, focusing on using rpi-image-gen as a tool rather than a framework.

2. Add Debian/Ubuntu amd64 capability
Unsupported or not, amd64 being the prominent workstation architecture, cross-building works. Instructions and automation will make the project more approachable. Just tell me where this goes - dependency install script or documentation?

3. Add build environment requirements and limitations
I stumbled on the fact that building on RPI3 was not possible at all, solid emulation is not readily available, and targeting below rpi4 is unclear. A few simple documentation updates here and there will clarify things.

If (any of) these are desired, I will create PR(s) targeting this issue.

johanehnberg:
## WIP tasks
- [x] Set the file location and if in root add a link in README or in doc add it to the ToC accordingly
- [x] Augment the guide with something common and simple yet descriptive. For example defining a one-liner command in the config that runs on target to install camera/AP/sensor. Settled on SSH and wifi.
- [x] Fix code snippet definitions sh->yaml for yaml files
- [x] Test everything before removing WIP

## Ideas for minimizing caveats and guide length outside the scope of this PR:
- Turn off PMAP for rpi3 and older
- Add layers for rpi1 rpi2 rpi0
- Move cross-build environment installs to install_deps.sh
- Add `raspberrypi.gpg.key` to keydir instead
- #208 
- #210 

Closes:  #206


learmj:
Hi @johanehnberg Let's just keep it all in one 'getting started' doc. Myself and others plus external / community will review it when it has content.

johanehnberg:
@learmj OK. If the general structure of the document sits well with the project, all I need is a good, common case to use in the example. Preferably something that helps people install their own app/assets/file and run a command, but the working example only needing one such line each.

learmj:
> @learmj OK. If the general structure of the document sits well with the project, all I need is a good, common case to use in the example. Preferably something that helps people install their own app/assets/file and run a command, but the working example only needing one such line each.

Sounds good. If you're basing it on your own use / experience, that would serve others well.

johanehnberg:
I added wifi station config as the simplest use case that we also have. It seems to have been requested so this may work as a stop-gap solution at least until a layer may be created. Considering the simplicity, I created #208 to allow streamlining much of this.

johanehnberg:
Ready for review. The guide is a bit longer than what I aimed for, but as the tool matures and built-in layers get added, it can be shortened. The contents are now such that would have saved me a lot of hours when I started.

getting_started.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+=== Project setup
+
+Let's set up a simple project that includes a configuration file and a script to run installation.
+
+[,sh]
+----
+mkdir -p myproject/bdebstrap
+echo -e 'ClientAliveInterval 60' > myproject/sshdkeepalive.conf
+echo -e '#!/bin/bash\n\nset -eu\n\nchroot $1 sh -c "apt-get -y install network-manager"\nchroot $1 sh -c "nmcli --offline connection add type wifi ssid $SSID wifi-sec.psk $PSK wifi-sec.key-mgmt wpa-psk > /etc/NetworkManager/system-connections/$SSID.nmconnection"\nchroot $1 sh -c "chmod 600 /etc/NetworkManager/system-connections/$SSID.nmconnection"' > myproject/bdebstrap/customize90-myscript
```
jaysephjw: had to populate `bdebstrap` first

getting_started.adoc
```diff
@@ -0,0 +1,123 @@
...
+  user1pass: Fo0bar!!
+
+image:
+  layer: image-rpios
+  boot_part_size: 128m
+  root_part_size: 1920m
+  assetdir: $(pwd)/myproject/
+
+device:
+  user1: myuser
```
jaysephjw:
```suggestion
```

duplicate `device` section

getting_started.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+device:
+  layer: rpi5
+----
+
+Building this will produce the specific image for the device you are deploying to location1 at the time.
+
+== Build Your Image
+
+Once your image configuration is ready, you can build everything with specifying just one configuration file. `customize90-wifi` will be automatically included based on its name, location and executability.
```
jaysephjw: What is `customize90-wifi`? maybe not used anymore

jaysephjw (commented):
Stumbled across this as I was looking at the project for the first time - and found it quite helpful!

Just a couple comments as I tried it.

learmj:
Hi @stu-spp @martinn-ronin
As users who have previously indicated a preference for some intro documentation, how does this look to you?

Hi @johanehnberg I'd prefer to wait on getting some user / community feedback before merging. Thanks again for your efforts.

getting_started.adoc
johanehnberg: The second `echo` should do that as it is, but let me know if something else related was needed?

getting_started.adoc
johanehnberg: Thanks, fixed.

getting_started.adoc
johanehnberg: Thanks, fixed to `customize90-myscript`

johanehnberg:
Thanks @jaysephjw for the feedback! All fixed.

stu-spp:
This looks great! :heart:

getting_started.adoc
```diff
@@ -0,0 +1,120 @@
...
+* The build target can be any Raspberry Pi, but all features are available only on RPi4 or later
+- When building for older targets, some layers will not work
+
+== Building on Ubuntu or Debian amd64
+
+While not officially supported, cross building works well on a recent Debian or Ubuntu amd64. In lieu of the quick start, use the following:
+
+[,sh]
+----
+sudo apt-get -q -y --no-install-recommends install binfmt-support qemu-user-static dirmngr slirp4netns quilt parted debootstrap zerofree libcap2-bin libarchive-tools xxd file kmod bc pigz arch-test debian-archive-keyring wget git
```
learmj: This dependency list appears to have been copied from pi-gen. Only a small number of these (host) deps are required, eg binfmt, qemu, debian-archive-keyring, git.

getting_started.adoc
```diff
@@ -0,0 +1,120 @@
...
+git clone https://github.com/raspberrypi/rpi-image-gen.git
+cd rpi-image-gen
+sudo ./install_deps.sh
+wget -O keydir/raspberrypi.gpg.key https://archive.raspberrypi.com/debian/raspberrypi.gpg.key
```
learmj: Not required because rpi-image-gen bootstraps the host apt with its own keydir

getting_started.adoc
johanehnberg: Actually combined from a few blog posts about `rpi-image-gen`, in the absense of the current doc. Good if we can reduce the list and retest.

getting_started.adoc
```diff
@@ -0,0 +1,120 @@
...
+----
+
+This will:
+
+* Install cross-build and further host dependencies
```
learmj: There's actually no 'cross-build' tools installed here. All this is doing is installing native dependencies if you want to perform foreign dpkg operations. For example, binfmt and qemu are not needed if you want to use rpi-image-gen to generate a native image for your host.

getting_started.adoc
```diff
@@ -0,0 +1,120 @@
...
+* Install Debian keyring from package on Ubuntu
+* Add Raspberry Pi key to your `rpi-image-gen` keys
```
learmj: As above

getting_started.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+== Define Your Image
+
+We will build three images starting with a single image for a single device. Then we use includes to show how you can define multiple images to manage your whole landscape. In this example, we will assume you have various RPi generations deployed across different locations serving as simple SSH servers on wifi.
```
learmj: I think only the final (3rd) configuration will create a bootable image, as the first two have no device layer.
Is the guidance such that only the third configuration file will be used to build an image?

getting_started.adoc
johanehnberg: Note the heading here is 'Building on Ubuntu or Debian amd64'. At the time of creating this PR, this key had to be added manually. Has that changed since?

getting_started.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+device:
+  hostname: location1
+
+env:
+  SSID: localssid
+  PSK: localpsk
+----
+
+This will define an image with a hostname specific to the location and override the timezone. Again, you can build this and it will work if your device is compatible with the default target of `rpi-image-gen` at the time.
```
learmj: You can build without a device layer specified, but I'm pretty sure it won't boot on anything (no kernel or firmware etc installed).

getting_started.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+Building this will produce the specific image for the device you are deploying to location1 at the time.
+
+== Build Your Image
+
+Once your image configuration is ready, you can build everything with specifying just one configuration file. `customize90-myscript` will be automatically included based on its name, executability and location in relation to the `-S` source directory flag.
+
+[,sh]
+----
+./rpi-image-gen build -S myproject/ -c myproject/myimage.yaml
```
learmj, line 118: As far as I can tell, this is the only build invocation that will produce a bootable image for a device. Is this the intention?

learmj (changes requested):

getting_started.adoc
johanehnberg: Again, this heading's context is amd64 building for arm64. It is not cross-compiling as such, but rather emulating for any system operations, so 'cross-build' seems the correct term to me. I don't mind it being called 'foreign dpkg operations' if that sits better with project terminology, though.

getting_started.adoc
learmj: The key used to authenticate rpi packages is present in the built-in keydir:
https://github.com/raspberrypi/rpi-image-gen/tree/master/keydir

This and the host keyrings are installed to a temporary keydir (`${@WORKROOT}/keys`), and this is passed to the host apt to use for the bootstrap:

https://github.com/raspberrypi/rpi-image-gen/blob/master/layer/base/sys-build-base.yaml#L26
https://github.com/raspberrypi/rpi-image-gen/blob/master/rpi-image-gen#L313

getting_started.adoc
johanehnberg: This is already mentioned in the doc. All of them create a bootable image since there are defaults.

getting_started.adoc
johanehnberg: Yes. However, it could also be repeated for each section if that improves readability.

getting_started.adoc
johanehnberg: Is your suggestion to install the key on the host instead?

I tried to retest to answer the question if this changed in the meantime. Currently the build fails on an apt 404 (potentially due to changes in trixie.source in the templates). I can check back once that is working again.

It would seem this PR is suffering from being open too long.

getting_started.adoc
johanehnberg: On retest now, the default seems to have been removed in the meantime, giving a 

`Error: Error: Layer 'image-base' requires provider 'device' but no layer in the dependency chain provides it`

I will add this (and any other necessary changes) once the overall build workflow is working again.

getting_started.adoc
learmj: While 'cross' is often used loosely in contexts like this, technical contexts typically use it to define _generating_ binaries for one architecture on a machine of a different architecture. What you're referring to here is different because it specifically relates to execution of non-native (eg arm64) binaries on an amd64 host, ie being able to invoke dpkg install of arm64 binaries in a chroot on an amd64 host, so the binary is being run via emulation.

I don't want to get too wrapped up in this :) but I just wanted to make the distinction clear, especially since rpi-image-gen does support a 'proper' cross build env set up, eg via toolchain layers like [this one](https://raspberrypi.github.io/rpi-image-gen/layer/arm64-toolchain-cross.html) used by [gnu-common](https://raspberrypi.github.io/rpi-image-gen/layer/gnu-common.html) for kernel cross-compilation.

I think _multiarch_ or _foreign dpkg operations_ or _non-native execution_ is probably more appropriate that cross-build here.

getting_started.adoc
johanehnberg: I did a retest again and found that the error was caused by some forward update breakage. I just re-cloned and it works again. I also found the key is now present, so removing that row.

getting_started.adoc
johanehnberg: Changed in the next commit.

getting_started.adoc
johanehnberg: Updated in next commit.

getting_started.adoc
johanehnberg: Updated.

getting_started.adoc
johanehnberg: Updated with a config default.

getting_started.adoc
```diff
@@ -0,0 +1,119 @@
...
+
+device:
+  hostname: location1
+
+env:
+  SSID: localssid
+  PSK: localpsk
+----
+
+This will define an image with a hostname specific to the location and override the timezone. Again, you can build this and it will work if your device is compatible with the default target of `rpi-image-gen` at the time.
```
devopsbob: How does this change or override the timezone? I had used the following in my config.yaml:
```script
locale:
  default: en_US.UTF-8
  keyboard_keymap: us
  keyboard_layout: us
  timezone: America/Detroit
```

devopsbob (commented):
Mentions that it will override the timezone but not which timezone or how.

getting_started.adoc
johanehnberg: Thanks, good catch! It was a remnant from a previous iteration that focused on locale. Fixed now.

getting_started.adoc
johanehnberg: I now minimized to that list, we can instead add back if any errors pop up.

martinn-ronin:
I truly appreciate all the work going into this tool and the effort to make it easier for more people to get started. The new getting started guide already helped me significantly compared to where I was before.

That said, I am still struggling to get my specific builds working and I find it difficult to understand how all the layers, suites, distributions, and device configurations are intended to fit together when moving beyond the provided examples.

I also feel the documentation and examples currently jump from simple concepts to more advanced compositions quite quickly without enough intermediate step-by-step guidance. As a new adopter, I often find myself unsure whether an issue is caused by my environment, my configuration, missing layer relationships, or unsupported combinations.

I really want to adopt this over my current system because the architecture and direction of the project look extremely promising, but at the moment the learning curve still feels quite steep for someone trying to build custom images outside of the exact provided examples.

johanehnberg:
@martinn-ronin thanks for your input! For the scope of this getting started document, would you say it works as a good second step after the README? Were there any issues that you stumbled upon? Is there a need to extend some aspect of it, so as to bridge the gap to the technical documentation in https://raspberrypi.github.io/rpi-image-gen/execution/index.html ?

martinn-ronin:
@johanehnberg Honestly, I am still not entirely sure what I am doing wrong, or even fully sure I understand the intended workflow yet.

The new getting started guide definitely helped me get my environment set up correctly and helped me progress further than before, so it was absolutely useful. However, I still find there is a fairly large gap between getting the tool installed/running and understanding how to confidently build my own image configuration from scratch.

At the moment, I can build the provided examples successfully, but that was already roughly the point I had reached before. I still struggle to understand how to create a simple custom image for my own use case (for example a Trixie armhf image for an RPi4), or whether certain combinations are currently supported or expected to work differently.

Part of this is definitely my own experience level. I have limited Linux experience outside of basic bash scripting, and very little prior experience with image build systems. I also found a YouTube tutorial that helped explain some concepts well, but it was made before the recent restructuring of the project, so some workflows no longer seem to match the current structure of the tool.

I think what would help me most would be a slower, more incremental walkthrough aimed at users who are new to this style of image building. For example:

* Build a very basic image successfully
* Explain how device, distribution, suite, and image layers relate to each other
* Then progressively extend the image step by step:

  * adding a user
  * adding Wi-Fi
  * adding packages
  * customising boot behaviour
  * slimming the image down
  * eventually adding OTA/update functionality

Right now I often find myself unsure whether an issue is caused by:

* my environment
* my configuration
* unsupported layer combinations
* or simply misunderstanding the intended workflow

I completely understand that documenting all of this thoroughly is a large amount of work, and I do not mean this as criticism. I mainly wanted to provide feedback from the perspective of someone who is very interested in adopting the tool, but is still early in the learning process.

johanehnberg:
Thanks for your detailed description. In many ways, your situation is very much the one I am trying to ease with this PR contribution. I am not affiliated with RPi, just giving back to a project of value to us, as I found myself in your shoes not long ago. Here are some thoughts of what you could try. Let me know what sticks, and I will consider if that can somehow be baked into the guide.

1. Before working on your own configuration file, to "build a very basic image" (and as I've noted in many cases, to debug with minimal dependencies and other baggage): build with just some command line arguments. Something like this should already give you a bootable image:
`./rpi-image-gen build -c ./config/trixie-minbase.yaml -- IGconf_device_layer=rpi4`

2. Build your first own config from the above arguments (or reducing the example to the minimum you can use to test). Add one configuration value at a time when iterating and testing.

3. Use known good tagged releases rather than master. This project is still introducing breaking changes. Similarly, try new releases with a known good config.

4. Write your customization scripts in a "non-rpi-image-gen" format, i.e. so you can test them easily on a stock image, where possible.

johanehnberg:
@martinn-ronin I saw you also mentioned `armhf` and #168 . We too have a use case for full armhf. Let me know if you find out how. Adding the below to the config did not suffice, even if trixie-minbase advertises armhf (it still installs arm64).

```
mmdebstrap:
  architectures:
    - armhf
```

I'll add it to the guide if solved.

martinn-ronin:
@johanehnberg I have currently decided to move past the full armhf userspace requirement for now, as the only armhf component we specifically need is support for running an armhf .NET application. I am currently exploring whether the arm64 + armhf multiarch setup is sufficient for this use case, since from my understanding the `arm64 armhf` configuration provides armhf foreign architecture support rather than a fully native armhf userspace.

If this approach works, then my next goal is to move fully over to `rpi-image-gen` and replace our existing provisioning/setup script with fully reproducible baked images, which would be a much better long-term solution for us operationally.

Unfortunately I probably will not be able to help much with solving the pure armhf userspace side of things, as I am still very new to this ecosystem. My closest experience to image building before this has mostly been Docker images and Bash scripting.

learmj:
I must thank @johanehnberg again for the contribution, but also for the continued work to make sure it's scoped correctly to help onboard new users. I really appreciate the sustained effort.

> @martinn-ronin I saw you also mentioned `armhf` and #168 . We too have a use case for full armhf. Let me know if you find out how. Adding the below to the config did not suffice, even if trixie-minbase advertises armhf (it still installs arm64).
> 
> ```
> mmdebstrap:
>   architectures:
>     - armhf
> ```

I don't know if conflating config file syntax with a layer only YAML key (`mmdebstrap`) was intentional here. If so, it isn't the first time I've both mentioned in the same context - they are not the same. From a usability and user feedback perspective, while the config parser will allow it, having...
```
mmdebstrap:
  architectures:
    - armhf
```
...in a _config_ file won't do anything, so is essentially a silent fail. The syntax is layer file only:
See the [bdebstrap](https://manpages.debian.org/unstable/bdebstrap/bdebstrap.1.en.html#YAML_CONFIGURATION) man page.
I wonder if it's worth detecting `mmdebstrap` in the config parser and bailing out with an informative error.

While YAML is the unified file format for configuration files _and_ layer files, the two are different and serve different purposes.

johanehnberg:
I believe the armhf question would be good to cover from both angles relevant here. After all, it is one of the three key properties of the official images, too (the two others being image content and release). As such, it is becoming clear to me what drives people to this tool rather than official images: to take control over those three basics.

The first angle would be the config validation, which already does similar sanity checks. The other would be the documentation, which I can do if you point me in the right direction. It seems building with just `IGconf_layer_base=raspbian-trixie-armhf` is not sufficient:

```
==> parameter_assembly

PARAM
CFG IGconf_device_layer=rpi3
CFG IGconf_layer_base=raspbian-trixie-armhf

ENV
<snip>
MMDEBSTRAP
Loaded rpi-boot-firmware
Loaded rpi-linux-v8
Loaded rpi-generic64
Loaded raspbian-trixie-armhf
<snip>
E: Unable to locate package linux-image-rpi-v8
<snip>
E: mmdebstrap failed to run
```

Is the tool currently capable of building armhf using the built-in layers?

johanehnberg:
For completeness, without device, the image generation does not work. Currently the device layer forces the architecture. Perhaps one approach would be to create architecture property for devices for conditional layer dependencies? Or just parallel device layers for armhf?

martinn-ronin:
I just wanted to add some thoughts on how I personally see many users approaching and wanting to use this tool.

From my perspective, many people getting started with Raspberry Pi systems usually begin by selecting a device (`Pi 4`, `Pi 5`, `CM4`, etc.), combining it with off-the-shelf hardware, and then building a system around that. Even the official Raspberry Pi Imager follows a very similar workflow:
1. Select device
2. Select operating system/distribution
3. Select architecture
4. Configure initial setup/customization
5. Flash and iterate from there

I think many users then eventually discover `rpi-image-gen` once they reach the point where they want:
- reproducible systems
- automated provisioning
- reduced setup scripting
- custom images
- better long-term maintainability

However, I think the current learning curve is very high for users who are not already familiar with building fully custom Linux images or embedded systems. The technical documentation around layers is very detailed and powerful, but it appears more oriented toward users who already understand concepts like custom distributions, image composition, layered build systems, dependency graphs, and low-level Linux provisioning.

The technical documentation on using layers actually suggests what I think is the correct mental model already:
- pick the device
- pick the base image/distribution
- add layers/features
- configure variables
- build

But in practice, I found it difficult to translate that into a real-world workflow when starting from scratch.

Personally, I think a workflow like the following would feel much more approachable to newer users:

1. Select target hardware (`Pi 4`, `Pi 5`, etc.)
2. Select architecture (`arm64`, `armhf`, multiarch, etc.)
3. Select a base OS profile (`minimal`, `desktop`, `lite`, etc.)
4. Add optional functionality layers:
   - desktop
   - VNC
   - Raspberry Pi Connect
   - SSH
   - networking customization
   - packages
   - provisioning
   - custom applications
5. Configure deployment variables
6. Build image

I think the examples currently available make the layering and composition model somewhat difficult to follow initially, especially for users coming from more common Raspberry Pi workflows rather than embedded Linux development backgrounds.

This is not criticism of the project itself — I actually think the architecture and long-term potential of `rpi-image-gen` is extremely powerful. One of the main reasons I want to persevere with learning it is because I can already see how much more maintainable and reproducible this workflow could become compared to backing up SSDs or relying heavily on post-install setup scripts.

If I can properly understand and adopt this workflow, I can see it becoming useful across many future projects where deterministic image generation makes far more sense than manually configuring systems after deployment. In many ways, I personally see this kind of layered image composition workflow as being as powerful for operating system images as Docker was for application deployment and containerization.

learmj:
@johanehnberg 

> Is the tool currently capable of building armhf using the built-in layers?

It is, but it looks as though you haven't included the layer which adds the RPi apt repo (eg [this one](https://raspberrypi.github.io/rpi-image-gen/layer/rpi-debian-trixie.html)). Since `linux-image-rpi-v8` is hosted there, apt has no way to locate it, hence your error. Note that you won't find `linux-image-rpi-v8` in the rpi [armhf repo](https://archive.raspberrypi.com/debian/dists/trixie/main/binary-armhf/Packages) anyway - only `linux-image-rpi-v6`.

FWIW, in my own personal view, there is little need for the Raspbian repo nowadays, unless working with legacy devices. Even Pi3 can execute 64-bit. 

Be careful generalising on 'armhf', though...

Some time ago, people were building images based on Debian armel, which is debian's armv6 soft float architecture. Using armv6 hard float resulted in a modest performance boost, but Debian's armhf architecture is armv7 and so not compatible with all devices at the time. Raspbian was started as a project to rebuild all of Debian armhf to be armv6hf and that's still what it hosts today. So while it is 'armhf', it's actually armv6hf so is compatible with all (RPi) devices.
See [https://forums.raspberrypi.com/viewtopic.php?t=22090](https://forums.raspberrypi.com/viewtopic.php?t=22090)

These days, arm64 is the typical primary architecture and when listed first in the YAML under the `architectures` key, apt treats it as the native architecture [[1]](https://manpages.debian.org/unstable/bdebstrap/bdebstrap.1.en.html#mmdebstrap). Additional architectures can be specified after the primary and these are treated as _foreign_. You can boot a 64-bit kernel (eg linux-image-rpi-v8) on a device with an ARMv8 instruction set and still use 32-bit armv6hf userland with it, but I'm unsure how many actual use cases there are for such an environment these days (@martinn-ronin this is what you were describing in #168).

> For completeness, without device, the image generation does not work.

The term 'image' can be ambiguous. rpi-image-gen can generate a filesystem as the input for a container 'image' file but it does not require a device layer to do so. If you're referring to using the tool to create a bootable disk image file for a device's storage media, then yes - it obviously needs a device layer for that.

> Currently the device layer forces the architecture. 

I suppose it does, but it doesn't have to. It's probably only that way due to the inter-dependency between layers. If the usage of `X-Env-Layer-Requires` [[2]](https://raspberrypi.github.io/rpi-image-gen/layer/index.html#_x_env_metadata) was tighter, the dep chain could resolve back to a 'Debian Base' and that could set the primary (ie native) architecture of the chroot. Actually, this is something I contemplated doing some time ago (and still might). It would be trivial enough to utilise `X-Env-Layer-RequiresProvider` to force a configuration that mandated a 'debian-base', therefore that would set the primary/native architecture of the chroot rather than the device, but it kind of seems sensible for the device layer to set the native architecture.

Also note that the `rpi-debian-trixie` layer does not force an architecture - it just inherits from what is set by apt (which may of course have been from upstream layers). It might surprise you that we build lots of packages for different architectures, including [amd64](https://archive.raspberrypi.com/debian/dists/trixie/main/binary-amd64/Packages).

@martinn-ronin Thanks very much for taking the time to write up your thoughts and views. It's greatly appreciated. That kind of feedback helps us understand the spectrum of users who are considering / trying to adopt using the tool. Ultimately, it helps us understand more about our target audience.

johanehnberg:
As a first and perhaps only requisite step, to me it sounds like a single working armv6hf native build - akin to the official 32-bit images - can cover all the remaining use cases:

- Legacy devices (rpi0 rp1 rpi2)
- Armhf userland
- The raspbian-exclusive software and rasbbian-exclusive patched versions of common software

That would keep the effort required on "legacy" to a minimum. Beyond that, the demand can likely only be for more optimized (zram, armv7hf?) or specific (arm64 kernel + armhf userland, other storage types than SD) approaches for already covered use cases, so further effort have much diminished returns.

I started looking into that now. However, I could not find any way to build with the existing layers, so I created new ones to introduce a start on being able to create actual native armhf builds. I got to the point that the build creates an SD image that boots on armv7l rather than arm64. The boot stops on looking for init, and the build stops post-image on IDP validation. @learmj would you like a draft PR with this proof-of-concept or just an issue?

learmj:
I'm not really sure. My gut says no (thanks). I don't know how many users would legitimately expect to be able to use rpi-image-gen to create bootable images for their rpi0, rp1, or rpi2 devices. In all likelihood, if those devices are operational and serving a purpose today, there would be questionable benefit in effectively reflashing with a totally different image and setting it back up again.

johanehnberg:
The use cases go beyond legacy devices. And it also raises the question, why are the raspbian, kernel 6 - 7l and related resources already included in the repository?

In any case, planned obsolescence does not resonate well with us, while justified maintenance burden considerations do. rpi1 through rpi3 still beat rpi4 and rpi5 in low power ethernet environments, and I am sure there are other cases like industrial integrations. RPi stands out it the SBC world with its good support.

learmj:
I added the raspbian repo a long time ago to keep parity with pi-gen - the tool that generates the official RPiOS images. There is no kernel 6 support in the repo.

> planned obsolescence does not resonate well with us

The practice of purposefully limiting the supply or durability of products in order to sell more (new) products is not the same as being [edit] **un**able to sustain high-volume manufacturing of a product.

Ref: [[1]](https://www.raspberrypi.com/news/raspberry-pis-commitment-to-longevity-a-sustainable-advantage/) and [[2]](https://www.raspberrypi.com/documentation/computers/raspberry-pi.html)

You misunderstood my previous comment.

johanehnberg:
Thanks for the info. It is up to this project to decide on those strategic topics. For the current case, it is just about software support, not manufacturing. I will wait to hear the outcome of the decision.

In the meantime, here's a look at the PoC: https://github.com/raspberrypi/rpi-image-gen/compare/master...johanehnberg:rpi-image-gen:armhf-native?expand=1

It makes for a good guide case at least.

learmj:
What's the feeling of the doc at the moment @johanehnberg ? I've deliberately not merged it so it can marinate and absorb comments.

johanehnberg:
I am a bit worried about the PR rot, but otherwise we are ready.

If the new wifi layer gets configurable SSID and PSK, the doc can again be adjusted to focus on something more beneficial with regard to the example case content. I think sticking to showing how to use your own customization script is still the approach. I might also slightly adjust it to cover that sequence of selections @martinn-ronin brought up, to the extent currently available.

A new example case could be running a [camera server](https://www.raspberrypi.com/documentation/computers/camera_software.html#tcp) in about three lines. I would still keep wifi configuration in the guide though.

learmj:
> If the new wifi layer gets configurable SSID and PSK

Some colleagues mentioned this earlier but I was not so enthusiastic. Having a user create their own profile file means rpi-image-gen is strictly 'hands off'. Plus, if there were any particular wlan settings needed, they would have to go in the file anyway. Providing an interface that supports generation of a profile file based on simple SSID + passphrase variables is too simplistic. X-Env Conflicts could be used to prevent build-time configuration conflicts with a profile file, but such a scheme seems overly complicated, not to mention that writing the profile file is supplicant specific.

johanehnberg:
I find 96% of environments we deploy in use PSK. The remainder are 3% open captive and 1% radius. That includes enterprise settings. Typically organizations just use VPN where it matters and most seem to consider their own wifi just another external/unsecured internet connection (the radius case being the exception).

What I can do if we really do not want to abstract SSID + PSK from iwd is to remain on the wifi use case in the doc, just updating it to show how to feed those parameters into the iwd profile using the same env as now.

learmj:
> What I can do if we really do not want to abstract SSID + PSK from iwd is to remain on the wifi use case in the doc, just updating it to show how to feed those parameters into the iwd profile using the same env as now.

#253 contains info of how to set a build-time profile. It will be in the doc page for the [iwd layer](https://raspberrypi.github.io/rpi-image-gen/layer/iwd.html) once merged and pages are generated.

learmj (approved):
LGTM and thanks again for taking the time to write this up.

The wlan connection activation via custom script is a really nice example. It demonstrates propagation of variables from a central config file which can then be inherited across different downstream device builds.

learmj:
I'm happy to merge this today if that's ok with you?

johanehnberg:
TLDR: Don't merge just yet, this needs a bit of aligning to address the wifi use case canonically.

I am currently working on switching the doc to use the iwd layer. I am reminded of one of my main gripes for orchestrating with iwd: the profile file name being the SSID. This is closely followed by having the config in `/var/lib/iwd/` rather than `/etc/`. It really conveys the message not to use it directly but through networkmanager or `iwctl`. Combined with the layer and profile approach, it accumulates to very inelegant code no matter how I turn it, since we need to activate iwd with a profile in the config:

As it is, either we need a separate profile file for each device or network, or some hacky `sed -i s/REPLACETHIS/$realpsk/ boilerplate.psk; mv boilerplate.psk \"$realssid.psk\"` which I cannot be sure runs after the boilerplate is present due to #210. If we want to benefit from the abstraction here, one somewhat elegant approach would be to allow an array of iwd profiles. This would help mobile RPis and add thus value with the layer.

That said, I'd simply `echo -e "[Security]\nPassphrase=${PSK}" > "$1/var/lib/iwd/${SSID}.psk"` which is elegant enough but not canonical `rpi-image-gen` so we should avoid that in a guide.

On a side note related to #212 , I also tested networkmanager + iwd and standalone iwd for AP use cases. It simply is not there yet on trixie. Device support is lacking, with many USB wifi dongles simply not entering AP mode at all (while with wpasupplicant they do).

learmj:
Understood.
I was going to investigate NM interop with iwd as I'm curious, and also to see how the landscape is in Trixie.
I take your point about external device support. It's relevant, but hopefully transitory, although I'm less hopeful regarding old dongles/devices. wpa_supplicant does carry a lot of old baggage and iwd travels lighter.

I'd say that auto-generation of an iwd profile as you suggest is reasonable, but an array of profiles would be more useful.
Re: hook ordering, you could always utilise an overlay in your SRCROOT. Overlays are applied first before the hooks run, so you could keep a boilerplate somewhere then cleanup after. See [here](https://github.com/raspberrypi/rpi-image-gen/blob/master/bin/runner#L218).

fyi after today, things will be quiet for a couple of weeks.

johanehnberg:
I pushed an update.

There are several things the guide does that add guide value that make sense to keep, that would not happen with changing to iwd profiles:
- Running something inside the chroot
- Leveraging env variables

iwd is not automatically picked up by networkmanager anyway, rather it uses wpasupplicant as default, so the guide remains smooth in relation to automatically included packages.

So instead, I just define the use case as wanting to rely on networkmanager. To me, that is a very solid case as argued above :smile: 

Ready to merge. The doc can then evolve with the project.

learmj (approved):

PR #215, opened 2026-03-27
cryptroot: LUKS disk encryption provisioning for initramfs-tools and dracut

learmj:
Implements production-ready LUKS cryptroot support driven by the image PMAP, covering both initramfs-tools and dracut initramfs generators across `image-rpios` and `image-rota` layers.

Three unlock schemes are supported:                                                                                                                 
                  
1.  initramfs-tools (custom) - single local-top script handles unlock. Does not use /etc/crypttab.
2. initramfs-tools (Debian) - standard Debian unlock via /etc/crypttab and a keyscript.
3. dracut (systemd) - `systemd-cryptsetup` driven by /etc/crypttab and a non-interactive passphrase agent.
                                                                                                                                                      
All schemes derive a hardware-bound, per-device passphrase via OTP HMAC (rpi-fw-crypto).

### Summary

- bin/pmap --cryptvars - new flag emitting POSIX shell variables for each LUKS2 container in the PMAP
- customize50-cryptroot - new top-level hook, detects LUKS volumes, selects backend, dispatches to installer
- customize60-dracut - new hook, equivalent to customize60-initramfs-tools (renamed)
- Initramfs overlay precedence fixed to base < image < device for both backends
- Slot-mapper dracut support: new 90rpi-ab-root module with pre-mount root-presence check
- eMMC write-protect: dracut module added using pre-pivot hook
- image-rpios - UUID/PMAP generation moved into chroot context (bdebstrap hook) so UUIDs are available at initramfs generation time


scripts/device/cryptroot/dracut/dracut.conf.d/rpi-cryptroot.conf
```diff
@@ -0,0 +1,12 @@
+add_dracutmodules+=" crypt dm "
+omit_dracutmodules+=" network nfs cifs mdraid dmraid "
+force_drivers+=" dm-mod dm-crypt af_alg algif_skcipher libaes aes_generic aes-arm64 libpoly1305 nhpoly1305 adiantum libchacha chacha-neon chacha_generic "
```
tdewey-rpi, line 3: Including the full range of unlock modules seems extraneous, when you should know apriori what cipher is being selected?

scripts/device/cryptroot/initramfs-tools/hooks/rpi-cryptroot-install
```diff
@@ -0,0 +1,58 @@
...
+copy_exec /usr/sbin/cryptsetup
+copy_exec /usr/bin/rpi-fw-crypto
+copy_exec /usr/bin/block-device-id
+
+# https://bugs.launchpad.net/ubuntu/+source/initramfs-tools/+bug/1958594
+for so in $(ldconfig -p | sed -nr 's/^\s*libgcc_s\.so.*=>\s*//p'); do
+   copy_exec "$so"
+done
+
+for module in $(find /usr/lib/modules/$version \
```
tdewey-rpi, line 28: Likewise, this could probably be specialised to select only the modules required for a given cipher choice?

scripts/device/cryptroot/dracut/dracut.conf.d/rpi-cryptroot.conf
learmj: Indeed, but this will necessitate new build-time device settings (likely via a new 'idp layer') which will require to be translated and relayed downstream to dracut and initramfs-tools paths in order for those hooks to install the appropriate cipher drivers, etc.
IMHO that work shouldn't block this PR, but it definitely does need doing.

scripts/device/cryptroot/initramfs-tools/hooks/rpi-cryptroot-install
learmj: As above.

image/gpt/ab_userdata/device/emmc/dracut.modules.d/90rpi-image/emmc-wp.sh
```diff
@@ -0,0 +1,37 @@
...
+WP_NUM_GRP=$(( (PROT_SZ_KB + WP_GRP_SZ_KB - 1) / WP_GRP_SZ_KB))
+
+# Num blocks to protect region
+WP_NUM_BLK=$((WP_NUM_GRP * WP_GRP_SZ_KB * 2))
+
+echo "WP $ERASE_GRP_SZ $WP_GRP_SZ $WP_GRP_SZ_KB $WP_NUM_GRP $WP_NUM_BLK"
+
+# Device geom dictates the min size that can be protected. The protection
+# applied must not overrun.
+WP_BYTES=$((WP_NUM_BLK * 512))
```
tdewey-rpi, line 31: Shouldn't this query the device for the physical block size? In particular, I'm mindful of 4k-native devices (I don't believe we have one as eMMC today, but who can predict the future?)

learmj: This script isn't new and was only touched in this change-set as a result of shuffling around initramfs assets, but it's more than that...
The range that gets protected is a combination of eMMC attributes read from the device (run-time), and logical sector size of the storage media and partition alignment size (build-time).

The logical sector size of the intended storage device is already included in the build params today (`IGconf_device_sector_size` see [here](https://raspberrypi.github.io/rpi-image-gen/layer/device-base.html).
The calculation of which to apply the WP is hard coded to assume 512 logical sector size, but `IGconf_device_sector_size` could be propagated.

However, the partition alignment size is currently not specified anywhere and is instead hardcoded as a result of  8afbaaa06e8a4f25313c04b8ba9016af3642b060 (original ref https://github.com/RPi-Distro/pi-gen/pull/833). The partition alignment should be exposed in a config variable via a layer, then propagated downstream to consumers like this script.

So to do the WP properly, partition alignment needs adding to a device/image layer and propagating, along with `IGconf_device_sector_size`, so the initramfs generation can use them. IMHO that's also something for a follow-up commit.

Lastly, it's not great having two copies of this script either, but the initramfs-tools version needs the header whereas the dracut script does not. They could be dynamically generated but that seems a little obfuscated.

roliver-rpi:
lock_device_private_key should probably be in the config.txt template.

learmj:
> lock_device_private_key should probably be in the config.txt template.

Yes, it should be.

scripts/device/cryptroot/initramfs-tools/hooks/rpi-cryptroot-install
roliver-rpi: For this particular instance, it might be worth considering the 'rpi-modcopy' script that rpi-sb-provisioner / pi-gen-micro uses to copy only the relevant modules.

learmj:
> lock_device_private_key should probably be in the config.txt template.

Done in add-on commit.

scripts/device/cryptroot/initramfs-tools/hooks/rpi-cryptroot-install
learmj: It would be a good solution, but I think using it would restrict builds to Raspberry Pi OS machines (or at least machines which are able to apt install from the Raspberry Pi repo). I'm trying to keep the tool flexible across build hosts, so unfortunately this is a no-can-do.

Issue #213, opened 2026-03-23
Invalid argument order on configure-splash in splash-screen.yaml


EmilienBINET:
If the `rpi-splash-screen` layer is used but with modified parameter, the image building process fails.

e.g. when **skip_image_checks = y** or **update_cmdline = n**, then `/usr/bin/configure-splash` is called with `$TEMP_IMAGE` as last argument.
The `configure-splash` tool specifies in its help `Usage: $0 <splash-image> [options]` and it is confirmed in its code with `IMAGE_PATH="$1"`.

So following the lines in the `layer/rpi/device/splash-screen/splash-screen.yaml` file:
```sh
      # Build command arguments based on boolean config values
      if [ "$IGconf_splash_skip_image_checks" = "y" ] && [ "$IGconf_splash_update_cmdline" != "y" ]; then
        chroot $1 /usr/bin/configure-splash --skip-image-checks --no-cmdline $TEMP_IMAGE
      elif [ "$IGconf_splash_skip_image_checks" = "y" ]; then
        chroot $1 /usr/bin/configure-splash --skip-image-checks $TEMP_IMAGE
      elif [ "$IGconf_splash_update_cmdline" != "y" ]; then
        chroot $1 /usr/bin/configure-splash --no-cmdline $TEMP_IMAGE
      else
        chroot $1 /usr/bin/configure-splash $TEMP_IMAGE
      fi
```
should become:
```sh
      # Build command arguments based on boolean config values
      if [ "$IGconf_splash_skip_image_checks" = "y" ] && [ "$IGconf_splash_update_cmdline" != "y" ]; then
        chroot $1 /usr/bin/configure-splash $TEMP_IMAGE --skip-image-checks --no-cmdline
      elif [ "$IGconf_splash_skip_image_checks" = "y" ]; then
        chroot $1 /usr/bin/configure-splash $TEMP_IMAGE --skip-image-checks
      elif [ "$IGconf_splash_update_cmdline" != "y" ]; then
        chroot $1 /usr/bin/configure-splash $TEMP_IMAGE --no-cmdline
      else
        chroot $1 /usr/bin/configure-splash $TEMP_IMAGE
      fi
```
to make it work.

learmj:
Thanks. That's a straight up bug. I don't think the splash layer received much testing verification before it was merged. It should have. I'll fix this in the next round of clean up.

PR #217, opened 2026-03-31
Fix path to EXAMPLE-splash.sh

lochel:


learmj:
How would you feel about including a fix for #213 in this PR?

learmj:
Merged. Thank you.

PR #218, opened 2026-03-31
Deployable .tar.zst to integrate with rpi-sb-provisioner

tdewey-rpi:
To simplify ingesting rpi-image-gen artefacts in rpi-sb-provisioner, generate .tar.zst archives that include the IDP json file and the various simg files.

### DIFF
--- a/layer/base/deploy-base.yaml
+++ b/layer/base/deploy-base.yaml
@@ -10,7 +10,7 @@
 # X-Env-VarRequires: IGconf_sys_workroot,IGconf_artefact_version
 # X-Env-VarRequires-Valid: string,string
 #
-# X-Env-Var-compression: none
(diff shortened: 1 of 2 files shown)


depends
```diff
@@ -20,6 +20,7 @@ build:rsync
 build:curl
 build:mtools
 build:pv
+build:zstd
```
learmj: Rather than this, it may be better to just always build zstd and install into the mini-sysroot. That way it's built with `-march=native`, and testing on a pi5 has shown that version to be quicker than Debian's for compression (~ 13%).
Might be better to just set `layer/base/deploy-base.yaml` `X-Env-Var-compression` = `zstd` and let the trigger build it.

learmj (approved):

PR #221, opened 2026-04-10
Layer overlay support and hook fixes

learmj:
- Add support for layers to define their own Filesystem Overlay. These overlays are applied in layer build order by `bin/runner` at the `customize` phase.
- Two bug fixes in hook execution.

The infrastructure added to support Layer Overlays lends itself towards being able to support a 'Yocto style' per-recipe task  where layers could participate in all build operations. Because `bin/runner` is now 'layer aware', it's possible to hook up things like per-layer hooks outside of what can be accomplished in the layer YAML itself.

### DIFF
--- a/bin/runner
+++ b/bin/runner
@@ -11,8 +11,9 @@ add_dir() {
    fi
 }
 
-phase="$1"
-shift 1
+PHASE="$1"
+PLAN="$2"
+shift 2
 
 
 # Script bundles in these dirs are executed in this order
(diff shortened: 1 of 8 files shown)


devopsbob:
Random thoughts: Could use a nice sequence diagram or c4 model description. Seems like layers and packages dependencies are sprawling across tooling spaces. ie. managing dynamic orders of executions depending on what is putting objects on the filesystem and when. decomposing that what and whens in c4 component model and building the plan/layers in a sequence.

PR #223, opened 2026-04-15
AB: boot-time robustness improvements

learmj:
Previously, only the boot and system slot components were qualified as being on the boot device.
Now, the AB layout uses GPT labels for all devices and ensures all mounted devices are boot device qualified.
This:
1. Promotes all mounted devices to exist under `/dev/disk/by-slot/`.
2. Ensures the backing block device can always be resolved (which is not true if using filesystem labels).
3. Ensures we unambiguously mount the devices we expect.

Also, ensure `nofail` is used to mount the bootconfig partition to help ensure the system boots to completion and doesn't stall. Doing so improves the chance of 'health management' tasks starting.

### DIFF
--- a/layer/rpi/device/ab-slots.yaml
+++ b/layer/rpi/device/ab-slots.yaml
@@ -25,7 +25,7 @@ mmdebstrap:
     - install -D -m 0755 "${IGconf_abslot_overlay}/bin/rpi-slot-label" "$1/usr/bin/rpi-slot-label"
     - install -D -m 0755 "${IGconf_abslot_overlay}/bin/rpi-slot-static" "$1/usr/bin/rpi-slot-static"
     - install -D -m 0755 "${IGconf_abslot_overlay}/bin/rpi-slot-tryboot" "$1/usr/bin/rpi-slot-tryboot"
-    - install -D -m 0644 "${IGconf_abslot_overlay}/udev/rules.d/slot.rules" "$1/etc/udev/rules.d/99-rpi-slot.rules"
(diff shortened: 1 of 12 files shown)


layer/rpi/device/ab-slots.yaml
```diff
@@ -25,7 +25,7 @@ mmdebstrap:
     - install -D -m 0755 "${IGconf_abslot_overlay}/bin/rpi-slot-label" "$1/usr/bin/rpi-slot-label"
     - install -D -m 0755 "${IGconf_abslot_overlay}/bin/rpi-slot-static" "$1/usr/bin/rpi-slot-static"
     - install -D -m 0755 "${IGconf_abslot_overlay}/bin/rpi-slot-tryboot" "$1/usr/bin/rpi-slot-tryboot"
-    - install -D -m 0644 "${IGconf_abslot_overlay}/udev/rules.d/slot.rules" "$1/etc/udev/rules.d/99-rpi-slot.rules"
+    - install -D -m 0644 "${IGconf_abslot_overlay}/udev/rules.d/slot.rules" "$1/etc/udev/rules.d/99-rpi-01-abslot.rules"
```
pelwell, line 28: Is this change covered by the commit message? It came as a surprise to me.

learmj: It isn't, but should've been. The rule has been renamed when installed  on the device so it's possible to have a deterministic lexical sort order for any rules that must follow it. For example, the new rule installed by the AB layer must sort after this one because it requires the (new) boot device marker. By using the 99-rpi-01 prefix here, it enables later rules to use a similar prefix therefore having a namespace to allow a guaranteed sort order, 
rather than just the 99-rpi-xxx which may not sort as expected.

PR #229, opened 2026-04-21
Support auth with device identities by default

clowder:
Include rpifwcrypto in the rpi-connect and rpi-connect-lite layers so images support authentication using the device-identity feature within Connect for Organisations by default.

### DIFF
--- a/examples/ota/README.md
+++ b/examples/ota/README.md
@@ -1,12 +1,28 @@
-Build an image that can be deployed remotely using Raspberry Pi Connect's
-experimental remote update capability.
+Build an image that can be deployed remotely using Raspberry Pi Connect's experimental remote update capability.
 
 This is a skeleton system that:
 
 * Installs a minimal base Trixie OS
-* Installs Raspberry Pi Connect Lite, configured with a given auth key
+* Installs Raspberry Pi Connect Lite
 * Installs and configures Raspberry Pi experimental OTA functionality
 
-```bash
-rpi-image-gen build -S ./examples/ota/ -c ota.yaml -- IGconf_connect_authkey=rpuak_XXX
-```
+A device can auto-signin with Raspberry Pi Connect on first boot in one of two ways:
+
+1. **Device identity** (Raspberry Pi Connect for Organisations only). The device authenticates using the unique identity in its OTP (one time programmable) memory. The identity must first be registered with your organisation via the [Raspberry Pi Connect Organisations Management API][management-api]. No credentials need to be embedded in the image:
+
+   ```bash
+   rpi-image-gen build -S ./examples/ota/ -c ota.yaml
+   ```
+
+2. **Auth key.** Embed a single-use auth key generated via your Raspberry Pi Connect dashboard:
+
+   ```bash
+   rpi-image-gen build -S ./examples/ota/ -c ota.yaml -- IGconf_connect_authkey=rpuak_XXX
+   ```
+
+If an auth key is embedded on a device that is also registered with an organisation, the auth key takes precedence.
+
+For further details, see the [Raspberry Pi Connect documentation][connect-docs].
+
+[management-api]: https://www.raspberrypi.com/documentation/services/connect.html#organisations-management-api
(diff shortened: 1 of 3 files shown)


mudge (approved):
Should we update the ota example to include a version that doesn't pass an auth key?

layer/rpi/device/services/rpi-connect-lite.yaml
```diff
@@ -28,6 +28,7 @@
 mmdebstrap:
   packages:
     - rpi-connect-lite
+    - rpifwcrypto
```
learmj, line 31: Is rpi-fw-crypto invoked by the Connect client in new versions? Just wondering why this commit is just tagging on an additional package dependency to be installed?

learmj (approved):

learmj:
BTW - rpi-image-gen now supports overlays per-layer, so in some cases it may be easier / more convenient to use that scheme if installing static content:
More info [here](https://raspberrypi.github.io/rpi-image-gen/execution/index.html#_overlays).

Currently, the Connect layers install dynamic content into the rootfs so this scheme is probably not useful in this application. Just thought I'd mention it.

layer/rpi/device/services/rpi-connect-lite.yaml
mudge: Yes, rpi-connect (and rpi-connect-lite) 2.11.0 will invoke `rpi-fw-crypto` to sign API requests when exchanging a device identity for an access token. As it is only available to users of Organisations, it is only a "suggested" dependency in the package for now.

clowder:
> Should we update the ota example to include a version that doesn't pass an auth key?

Good idea: I've pushed d7808ca56bf01107172ffcc640bcc30da6f91c59 to add something to the README.md.

layer/rpi/device/services/rpi-connect-lite.yaml
learmj: There's a lib if you wanted to dlopen / tryload that instead of shelling out (librpifwcrypto0).

examples/ota/README.md
```diff
@@ -4,9 +4,23 @@ experimental remote update capability.
 This is a skeleton system that:
 
 * Installs a minimal base Trixie OS
-* Installs Raspberry Pi Connect Lite, configured with a given auth key
+* Installs Raspberry Pi Connect Lite
 * Installs and configures Raspberry Pi experimental OTA functionality
 
+By default, the device authenticates with Raspberry Pi Connect on boot using
+the device identity in its OTP (one time programmable) memory. The device
```
mudge: Nitpick but we use "one-time programmable" as the definition of "OTP" in [our documentation](https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#otp-register-and-bit-definitions).

mudge (approved):

examples/ota/README.md
```diff
@@ -4,9 +4,23 @@ experimental remote update capability.
...
+identity must first be registered with your organisation via the [Raspberry
+Pi Connect Organisations Management API][management-api].
```
learmj: Should we use slightly different wording here since device identity auto-binding is Org only and orthogonal to `IGconf_connect_authkey`?

I know there is a clause below, but maybe it would be clearer if it was laid it out as two brief bullet points, eg...

> There are two ways in which a device can auto-register...
> 1) Discrete, single use auth key
> 2) Device unique identy
> 

What happens if a user supplies one when the device is already Org registered? Presumably the Org registration path would take precedence?

Feels like we should maybe redirect to a richer documentation set here, ie "for further details refer to ..."

clowder: Updated in 7f228e264e34fce567f8149b977edebdb5c6edbb, preview:
https://github.com/raspberrypi/rpi-image-gen/blob/7f228e264e34fce567f8149b977edebdb5c6edbb/examples/ota/README.md

Restructured as two numbered options with a pointer to the Connect documentation for further detail. Regarding precedence, the auth key wins. The assumption is that a user who put one on disk did so deliberately.

learmj:
LGTM Thanks

PR #233, opened 2026-04-24
Boot device qualification of all block devices, including LUKS volumes

learmj:
This change-set ensures all block devices that are part of the boot image, whether encrypted or not, are on the boot device. `image-rota` and `image-rpios` now create their own `/etc/crypttab` containing `by-slot` aliases for LUKS volumes requiring unlock with initramfs-tools and dracut support added for those initramfs generation frameworks. Usage of `by-slot` is widened to non-AB and denotes block devices which are special and boot-device qualified.

### DIFF
--- a/docs/layer/image-rota.html
+++ b/docs/layer/image-rota.html
@@ -123,7 +123,7 @@
     <div class="header">
         <h1>image-rota</h1>
         <span class="badge">image</span>
-        <span class="badge">v5.3.0</span>
+        <span class="badge">v5.4.0</span>
         <p>Immutable GPT A/B layout for rotational OTA updates,
  boot/system redundancy, and a shared persistent data partition.</p>
     </div>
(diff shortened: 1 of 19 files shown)


pelwell (commented):
There looked to be some near-duplication of files in this set, but I suspect it is just required boiler-plate. What I haven't seen is /etc/udev/rules.d/99-rpi-06-image-crypt.rules - is it missing?

learmj:
That file is auto-generated in order for it to have encrypted volume rules that are derived from the pmap. The image-rota later supports different pmaps, with the desired one being selected build time. The encrypted volume in each pmap has different attributes (eg label). The choice for the corresponding rules file is either for it to contain static information matching the pmap, or for it to be generated from the pmap (one source of truth).

pelwell:
Got it.

learmj:
Successfully OTA updated a device previously provisioned with LUKS. All good.

PR #236, opened 2026-04-29
Ship VideoCore udev rules in rpi-connect layers

clowder:
Minbase images don't include `raspberrypi-sys-mods`, so `/dev/vcio` stays `0600 root:root`. The non-root user is in the `video` group (via `rpi-user-credentials`) but without a udev rule the group has no permissions. As a result using `rpi-fw-crypto` fails with "Permission denied".

Pulling in `raspberrypi-sys-mods` would also fix this issue, but it brings a lot of baggage with it.

### DIFF
--- a/layer/rpi/device/services/rpi-connect-lite.yaml
+++ b/layer/rpi/device/services/rpi-connect-lite.yaml
@@ -31,6 +31,13 @@ mmdebstrap:
     - rpifwcrypto
   customize-hooks:
     - |-
+      cat > "$1/usr/lib/udev/rules.d/10-vc.rules" << 'UDEVRULES'
+      KERNEL=="vcio",     GROUP="video", MODE="0660"
+      KERNEL=="vchiq",    GROUP="video", MODE="0660"
+      SUBSYSTEM=="vc-sm", GROUP="video", MODE="0660"
+      KERNEL=="vcsm-cma", GROUP="video", MODE="0660"
+      UDEVRULES
+    - |-
       set -eu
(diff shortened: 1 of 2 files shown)


learmj:
I don't think shipping udev rules to fixup a device (`/dev/vcio`) group ownership and permissions belongs in this layer. 

As discussed, this is a wider issue and should be resolved in a structured and sensible way, such as an rpi device specific base layer using its overlay/hooks to apply expected/assumed group ownership and permissions to device nodes. While `raspberrypi-sys-mods` already does this, it comes with a lot of extra baggage (including additional pkg dependencies) that would otherwise not be needed in this particular case.

Also, in accordance with standard package install flow, if a program/service requires the user its running as to be a member of any groups, its postinst hook should add the user to those groups rather than relying on the base OS.

clowder:
> I don't think shipping udev rules to fixup a device (`/dev/vcio`) group ownership and permissions belongs in this layer.
> 
> As discussed, this is a wider issue and should be resolved in a structured and sensible way, such as an rpi device specific base layer using its overlay/hooks to apply expected/assumed group ownership and permissions to device nodes. While `raspberrypi-sys-mods` already does this, it comes with a lot of extra baggage (including additional pkg dependencies) that would otherwise not be needed in this particular case.

Agreed — as we discussed, I'll close this in favour of the rpi-device-base layer approach you proposed

> Also, in accordance with standard package install flow, if a program/service requires the user its running as to be a member of any groups, its postinst hook should add the user to those groups rather than relying on the base OS.

Worth noting that in our OTA examples the user is a member of the appropriate groups, it's just that those groups don't have the expected permissions to access the underlying hardware (which they would have on Raspberry Pi OS).

PR #240, opened 2026-05-07
IDP: Enable partition expansion in all PMAPs

learmj:
Enable expand-to-fit on applicable partitions in all provisioning maps so the provisioner extends the partition table entry to the end of the target device. For crypt PMAPs, expand-to-fit is set on both the encrypted volume (so its partition entry grows on disk) and the last partition within the volume (so the encapsulated partition entry reaches the end of the volume).


learmj:
fyi @tdewey-rpi

PR #242, opened 2026-05-11
Add new common rpi device and security policy layers

learmj:


learmj:
@clowder

clowder:
Ran through this with the OTA example and the unreleased rpi-sb-provisioner (with device identity support) and everything worked as expected. Connect was able to exchange the devices's identity for an access key using rpi-fw-crypto!

PR #250, opened 2026-05-18
layer (suite): Add wlan support to trixie minbase

learmj:
All downstream consumers of trixie-minbase will gain wireless LAN support. This includes the default build configuration targets and OTA example.

### DIFF
--- a/layer/suite/debian/trixie-minbase.yaml
+++ b/layer/suite/debian/trixie-minbase.yaml
@@ -1,8 +1,8 @@
 # METABEGIN
 # X-Env-Layer-Name: trixie-minbase
-# X-Env-Layer-Desc: Debian Trixie with apt, utils, wired networking, ssh.
+# X-Env-Layer-Desc: Debian Trixie with apt, utils, wired and wifi networking, ssh.
 # X-Env-Layer-Category: suite
-# X-Env-Layer-Version: 1.1.0
+# X-Env-Layer-Version: 2.0.0
 # X-Env-Layer-Requires: debian-trixie-arm64-multi,
 #  rpi-debian-trixie,
 #  rpi-misc-utils,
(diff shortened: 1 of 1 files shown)


learmj:
fyi @clowder @mudge @pelwell

pelwell:
Errm, is that all it takes?

learmj:
> Errm, is that all it takes?

Need to bake in AP auth credentials, but that's it. #246 merged last week has further details about how to do that.

pelwell (approved):

devopsbob:
Nice feature add along the recent adoc discussion adds but..does it silently fail if the psk file isn't present or inform in the logs wifi is enabling because I was going for zero trust least privilege network by not having it both wired and wireless active (or available) so adding the wireless I wanna know how to shut it off and not install those libraries...(I'll try the build weekend after holiday ;->)

learmj:
> does it silently fail if the psk file isn't present 

No. iwd will start, find no profiles, and sit idle scanning periodically. It won't attempt to connect to anything. The interface will be up but unassociated. This seems reasonable behaviour for a base image - iwd is present and functional, a profile can be dropped in at provisioning time or by the user, and iwd will pick it up and connect automatically without needing a restart.

> I wanna know how to shut it off and not install those libraries

The two additional layers (`wireless-regulatory` and `iwd`) only exist in one place to allow them to be included in two default build configurations (`config/trixie-minbase.yaml` and ` config/trixie-minbase-ab.yaml`), via `layer/suite/debian/trixie-minbase.yaml`. I'd expect custom image builders to be using their own set of layers where they have full control.

devopsbob:
ok. thanks for the clarification. The iwd scanning interval on my network is what I was keeping off. Yes, reasonable for a default setup. (I'll check what how it runs with config.txt dtoverlay=disable-wifi)

learmj:
I'd recommend you build your system from the required layers rather than using the default configs. That way you own the configuration. wlan0 was always available as an interface but nothing was managing it until now.

PR #254, opened 2026-05-21
examples/ota: Add wlan config section to readme

learmj:


examples/ota/README.md
```diff
@@ -6,6 +6,8 @@ This is a skeleton system that:
 * Installs Raspberry Pi Connect Lite
 * Installs and configures Raspberry Pi experimental OTA functionality
 
+One of the by-products of the build is `update.tar.zst` which can be deployed to the device via Raspberry Pi Connect to update it.
```
clowder: Worth putting in the whole path? In our future guide for sb-provisioner deploys we tell folks to look for `work/deploy-1.0.0/myapp-1.0.0-1.0.0.tar.zst`.

examples/ota/README.md
```diff
@@ -24,5 +26,24 @@ If an auth key is embedded on a device that is also registered with an organisat
...
+
+```yaml
+iwd:
+  profile: ${@SRCROOT}/My Network.psk
+
+ieee80211:
+  regdom: GB
+```
+
+Place `My Network.psk` directly under `examples/ota/` - rpi-image-gen sets that as its source directory for this example, so `${@SRCROOT}` resolves there. Set `ieee80211.regdom` to the ISO 3166-1 alpha-2 country code for your region.
```
clowder: I think it's worth replacing  "My Network" a placeholder to make it clearer the user should be providing the value.

````suggestion
```yaml
iwd:
  profile: ${@SRCROOT}/<SSID>.psk

ieee80211:
  regdom: GB
```

Create `<SSID>.psk` under `examples/ota/`, where `<SSID>` matches the name of your wireless network — iwd uses the filename to identify the network, so the `profile:` value must match. Set `ieee80211.regdom` to the ISO 3166-1 alpha-2 country code for your region.
````

**Review by clowder — COMMENTED** [review]
Couple minor things, otherwise this is a great quality of life improvement!

File: examples/ota/README.md (hunk 1 above)
Comment (learmj) [reply examples/ota/README.md]: Are you sure about that path? I don't see the slot pair update file (`update.tar.zst`) in the deploy dir.
The provisioner bundle is not an update file - it's the entire base system + IDP assets for provisioning.

Comment (clowder) [reply examples/ota/README.md]: I guess implied in my question is, I don't know. I guess my main worry is the build process outputs a lot of artefacts, so finding a file without the full path might be tricky for new users.

**Comment by learmj** [comment]
How does that look?

**Comment by learmj** [comment]
@stevedavies-rp fyi

File: examples/ota/README.md [hunk 3]
```diff
@@ -24,5 +26,30 @@ If an auth key is embedded on a device that is also registered with an organisat
 
 For further details, see the [Raspberry Pi Connect documentation][connect-docs].
 
+## Wireless Networking
+
+The minimal base Trixie OS enables wired and wireless networking by default. To have the device associate to a wireless AP automatically on boot using WPA2, pre-seed an iwd profile by adding the following to `ota.yaml`:
```
clowder: Stumbled on this getting it running:

```suggestion
The minimal base Trixie OS enables wired and wireless networking by default. To have the device associate to a wireless AP automatically on boot using WPA2, pre-seed an iwd profile by adding the following to `./examples/ota/config/ota.yaml`:
```

PR #262, opened 2026-06-10
layer: add NetworkManager support

learmj:
Two new layers:

network-manager: installs and enables NetworkManager with a slot-shared system-connections directory so network profiles survive A/B slot updates. Supports optional build-time connection provisioning via IGconf_nm_cmds.

network-manager-iwd: composition layer that configures NM to use iwd as the wifi backend with NM retaining control of auto-connect.

### DIFF
--- a/layer/net-misc/network-manager-iwd.yaml
+++ b/layer/net-misc/network-manager-iwd.yaml
@@ -1,0 +1,12 @@
+# METABEGIN
+# X-Env-Layer-Name: network-manager-iwd
+# X-Env-Layer-Desc: NetworkManager with iwd as the wifi backend, NM managing
+#  connections and auto-connect
+# X-Env-Layer-Version: 1.0.0
+# X-Env-Layer-Category: connectivity
+# X-Env-Layer-Requires: network-manager,iwd
+# METAEND
+---
+mmdebstrap:
+  customize-hooks:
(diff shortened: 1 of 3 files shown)


learmj:
Hi @johanehnberg based on our previous discussions, how do these look to you?

johanehnberg:
We might want to manage this `--offline` bug: https://redhat.atlassian.net/browse/RHEL-74255

By adding a `sed -i 's/Warning.*//'` or similar while piping the output.

The bug is fixed in 1.53, and trixie has 1.52.

johanehnberg:
Right now this would go into the example, since the cmds is essentially a script that we can do anything with.

Beyond that, I think the approach should work well.

learmj:
Yes, manipulation of the nmcli cmd output can be done inside the user's cmd file so it occurs before the output is written to a file in the chroot. For example, using this cmd file works:
```
$ cat nm.cmds 
# Testing
--offline connection add type wifi ifname wlan0 con-name SSID ssid SSID wifi-sec.auth-alg open wifi-sec.key-mgmt wpa-psk wifi-sec.psk PSK connection.autoconnect-retries 0 | grep -v '^Warning' > /etc/NetworkManager/system-connections/SSID.nmconnection
```
I'll update the description to specifically spell out that cmds in the file are executed by the shell, so pipes and redirection are supported.
I appreciate the info and your feedback. Thanks a lot.

PR #269, opened 2026-06-19
Refactor user credential layers and harden sudo group membership

learmj:
This series introduces two new base layers for user account management, makes user account provisioning generic, and hardens sudo group membership controls.

### New layers

User account variables were previously defined in `device-base` and provisioning performed by `rpi-user-credentials`. This series splits those concerns into a clean three-layer chain:

device-base -> device-user-admin -> device-user-credentials

- `device-user-credentials` declares all `IGconf_device_user1*` variables (extracted from `device-base`)
- `device-user-admin` performs all user account provisioning (renamed and relocated from `layer/rpi/user-credentials.yaml`)

`device-base` requires `device-user-admin`, so user account provisioning is automatic for all device builds.

### Configurable group membership

`IGconf_device_user1groups` replaces the hardcoded supplementary group list. Groups not present on the system are created as system groups before the user is added.

After group definition was moved to a variable, there was nothing RPi specific about the user account creation beyond a template, so drop that and use echo to install the `nopasswd` setting instead.

### Hardened sudo group membership

`sudo` is removed from the default group list, with group membership now added exclusively by the sudo hook when `user1sudo` is `passwd` or `nopasswd`. Declare a conflict on user1groups if sudo is present, avoiding sudo being introduced via that path.

### Breaking change

Layers requiring `rpi-user-credentials` must be updated to require `device-user-admin`.

### DIFF
--- a/docs/layer/bookworm-minbase.html
+++ b/docs/layer/bookworm-minbase.html
@@ -135,7 +135,6 @@
             <a href="rpi-misc-utils.html" class="dep-badge">rpi-misc-utils</a>
             <a href="rpi-essential-base.html" class="dep-badge">rpi-essential-base</a>
             <a href="rpi-misc-skel.html" class="dep-badge">rpi-misc-skel</a>
-            <a href="rpi-user-credentials.html" class="dep-badge">rpi-user-credentials</a>
             <a href="systemd-net-min.html" class="dep-badge">systemd-net-min</a>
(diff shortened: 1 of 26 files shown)


learmj:
@mudge @clowder The rpi-connect layers are touched by this change-set, but there's no functional change. The diff is simply pulling in a different layer for the `user1` credentials.

PR #273, opened 2026-06-25
storage-binder: support USB mass storage (USB-MSD) boot

ghollingworth:
The by-slot root resolver (rpi-bootdev-tag + storage-binder udev rules) only handled SD (boot-mode 1 -> mmcblk0) and NVMe (boot-mode 6 -> nvme0n1). When a Pi boots from USB mass storage the tagger hit the default case and exited, so no partition was tagged RPI_ONBOOTDEV and /dev/disk/by-slot/system was never created: the initramfs gives up waiting for the root device and drops to a shell.

Add USB-MSD support:
- rpi-bootdev-tag: handle boot-mode 4/5. The firmware does not name the USB device (and USB enumeration is non-deterministic), so discover the boot disk by content - the USB-attached disk carrying our vfat BOOT- labelled partition.
- storage-binder.rules: run sd* partitions through the tagger.
- rpi-bootdev-install: bundle blkid into the initramfs for the probe.

The label->by-slot rule (99-rpi-05-image) is already device-agnostic and needs no change.

Tested on a Pi 5 host: the tagger tags only the USB boot disk's partitions, and the full udev chain creates by-slot/{system,boot} -> the USB root/boot partitions. The exact USB boot-mode value(s) should be confirmed on hardware that boots from USB.

### DIFF
--- a/layer/rpi/device/storage-binder/initramfs-tools/hooks/rpi-bootdev-install
+++ b/layer/rpi/device/storage-binder/initramfs-tools/hooks/rpi-bootdev-install
@@ -8,4 +8,6 @@ esac
 
 copy_exec /usr/bin/od
 copy_exec /usr/bin/rpi-bootdev-tag
+# blkid is needed by rpi-bootdev-tag's USB boot-disk discovery (label/type probe).
+copy_exec /sbin/blkid 2>/dev/null || copy_exec /usr/sbin/blkid 2>/dev/null || true
 copy_file config /etc/udev/rules.d/99-rpi-00-bootdev.rules /etc/udev/rules.d/99-rpi-00-bootdev.rules
--- a/layer/rpi/device/storage-binder/bin/rpi-bootdev-tag
+++ b/layer/rpi/device/storage-binder/bin/rpi-bootdev-tag
@@ -1,11 +1,11 @@
 #!/bin/sh
 # shellcheck shell=sh
 
-set -u
-
 # Tag an incoming block device as boot-device-backed for udev IMPORT.
 # This script is boot critical and must be compatible with minimal shell
 # environments such as klibc, and be as POSIX compliant as possible.
+
+set -u
 
 die() {
    msg="ERROR: $*"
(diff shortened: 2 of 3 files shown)


learmj:
It would be nicer to walk sysfs or have a lighter-weight scheme instead of pulling `/sbin/blkid` into the initramfs.
Does it work if you use `readlink -f /dev/disk/by-partlabel/BOOT` instead?
Since this is only when operating in MSD mode we can be a bit more relaxed.

ghollingworth:
I've not done any further work on it...  I was just going to submit an issue, but this is a claude generated patch instead!

Customer assumed that inserting an SD card into a Pi 5 was the same as inserting a USB SD card reader.

timg236:
For reference, the bootloader will report the USB route string but physical topology isn’t stable with hot plugging.

Maybe most robust is a pair of device-type and GUID with GPT partitions if we have settled on that for OTA.

But there would be a small project for someone to update bootloader and start.elf to propagate selected GUID for partition. At least with rpi-image-gen we just set min firmware dependencies to ensure firmware has this feature.

learmj:
> Customer assumed that inserting an SD card into a Pi 5 was the same as inserting a USB SD card reader.

RPiOS isn't strict about this, but rpi-image-gen is (and I think with good reason). I did touch on the same topic a little while back - see [here](https://github.com/raspberrypi/rpi-sb-provisioner/issues/296#issuecomment-4517118912). My comment above is wrong - we do need to make sure that the boot disk is legit and as intended. I'm not convinced we can do that easily with USB MSD boot.

Using a GUID isn't a guaranteed way of ensuring booting from a particular device since it's an ID that  can easily be spoofed (see linked comment above). The disk needs to be correlated to hardware (somehow), eg could have two USB disks connected with the same label, GUID, etc.

layer/rpi/device/storage-binder/initramfs-tools/hooks/rpi-bootdev-install
```diff
@@ -8,4 +8,6 @@ esac
 
 copy_exec /usr/bin/od
 copy_exec /usr/bin/rpi-bootdev-tag
+# blkid is needed by rpi-bootdev-tag's USB boot-disk discovery (label/type probe).
+copy_exec /sbin/blkid 2>/dev/null || copy_exec /usr/sbin/blkid 2>/dev/null || true
```
learmj, line 12: Also, this layer supports initramfs generation using dracut. Therefore any binaries that additionally need to be copied into the initramfs to support new functionality must be installed when using dracut, too. If not, we deviate in functionality between different boot modes and initramfs providers.

learmj (changes requested):

ghollingworth:
Correlation between hardware and disks (if really required) is done through an encryption key stored in the OTP.  If it fails to decrypt then it should fail to boot.  If a customer complains that it doesn't boot if they plug in a new media source which has a valid image on it, then that's a "won't fix"

learmj:
OTP-backed encryption is a strong control for storage media substitution - agreed. Not disputing that. But the storage-binder check is complementary for two reasons that encryption doesn't address:

1. Availability, not just substitution. No attacker required: a USB device that enumerates before the real root disk and shares a filesystem UUID or GPT PARTLABEL etc causes a boot failure or worse, a silent wrong device mount. This is an operational hazard, but encryption doesn't help because the kernel has already resolved the wrong block device for root.
2. cmdline.txt is in the clear. The `root=` identifier in the vfat partition is unencrypted and readable by anyone with physical access. An attacker can duplicate that identifier on a rogue device to manipulate which device the initramfs uses for root. Best case with OTP encryption: a decryption failure and a reboot loop. But the attack surface exists and boot-disk binding prevents it earlier in the chain, before the root device is mounted.

storage-binder is part of what makes the boot chain reliable and deterministic, rather than dependent on decryption being the only gate. It's also worth noting that storage-binder operates below the encryption layer, ie it fires in early userspace before cryptsetup is invoked to unlock LUKS. USB can't be supported as-is because the hardware path that storage-binder relies on would be port-bound, not device-bound, ie there's no physical constraint that ties a specific device to that path across boots.

Note that rpi-sb-provisioner only supports sd, emmc and nvme storage - USB isn't a supported storage type.
https://github.com/raspberrypi/rpi-sb-provisioner/#step-3-configure-options

Issue #278, opened 2026-07-06
Feedback from using this tool in a container


vallahaye:
I've been using `rpi-image-gen` to generate golden images for my infrastructure node running on a Raspberry Pi 4 in my home lab. Since I don't have a spare Raspberry Pi to execute the toolchain, I've been trying to make it work inside a [Debian Trixie](https://hub.docker.com/_/debian) container on my Arch Linux host. Below is a list of the issues I encountered while doing so, along with fixes and ideas to improve this awesome project.

For context, this is the `Containerfile` I wrote:

```dockerfile
FROM docker.io/debian:trixie

ARG DEBIAN_FRONTEND=noninteractive

ENV RPI_IMAGE_GEN_VERSION=2.7.0

RUN <<'EORUN'
#!/usr/bin/env bash
set -xeuo pipefail
apt-get update
apt-get -y --no-install-recommends --no-install-suggests install ca-certificates git
git clone --branch="v${RPI_IMAGE_GEN_VERSION}" --depth=1 https://github.com/raspberrypi/rpi-image-gen.git /usr/share/rpi-image-gen
EORUN

WORKDIR /usr/share/rpi-image-gen
COPY patches/rpi-image-gen/*.patch patches/
COPY patches/bdebstrap/*.patch package/bdebstrap/patches/0.7.0/

RUN <<'EORUN'
#!/usr/bin/env bash
set -xeuo pipefail
git apply patches/*.patch
./install_deps.sh
EORUN

ENTRYPOINT ["./rpi-image-gen"]
CMD ["--help"]
```

## Issues with `install_deps.sh`

As `rpi-image-gen` is supposed to run directly on the host, [it verifies that the `binfmt_misc` filesystem is properly mounted](https://github.com/raspberrypi/rpi-image-gen/blob/v2.7.0/lib/dependencies.sh#L109-L115) when checking for dependencies on a non-ARM systems. This becomes problematic when the `install_deps.sh` script is executed inside a container, as `binfmt_misc` is expected to be configured on the host rather than within the container itself. As a result, requiring `binfmt_misc` to be mounted inside the container is unnecessary and prevents otherwise valid containerized setups from running successfully.

One solution is to relax the check by verifying only that the `binfmt_misc` filesystem is available, rather than requiring it to be mounted:

```patch
diff --git a/lib/dependencies.sh b/lib/dependencies.sh
index 3371a81..5052bf6 100644
--- a/lib/dependencies.sh
[...]

PR #280, opened 2026-07-09
Check if the binfmt_misc file system is supported, rather than mounted

vallahaye:
As discussed in #278.


learmj:
Thanks. I agree with this change, but I need to give it some more thought. I think it's definitely friendlier to container run-time environments and the test works correctly there, and also in non-container envs. What's not to like :)

learmj:
Thanks @vallahaye

PR #282, opened 2026-07-15
feat: traits - part 1

learmj:
This is the first part of a planned multi-stage submission (already tested end to end) which introduces a trait token registry to rpi-image-gen. The registry is a tree of DEB822 metadata files describing hardware and boot-mechanism facts (eg `hw:storage:nvme`, `hw:soc:bcm2712`, `boot:gpt`), parsed through the existing X-Env metadata/env/validator machinery. This opens the door to layers being able to understand much more about the underlying hardware platform that the build is targeting.

- `site/trait_registry.py` (new) - resolves the tree into semantic tokens, expands `Triggers:` (a trait activating others), validates `Requires:` (trait dependencies), and exposes the result via `rpi-image-gen config --trait`.
- `has()` / `not has()` added to `conditions.py`, so a condition can test trait presence anywhere the existing `IGconf_*` condition syntax is accepted. These helpers are not yet reachable from the layer engine.
- OEM source roots (ie `-S <path/to/custom/tree`) can add their own trait tokens and extend built-in namespaces via `Includes: `. See `examples/custom_layers/`.
- New `test/trait/` component with static DEB822 fixtures for lint and a Python-driven test for load/resolve (eg forward references, duplicate defs, etc).

We need to drop INI file syntax support from the config parser because it will not support how traits can be overridden from the config file.

## Scope

This is intentionally the first of several planned steps:

- **Traits are boolean only in this PR.** `X-Env-Trait-*-Valid` is hard-wired to `bool`.  Support of typed (eg, int/string/regex/enum) traits, reusing the same validator infrastructure, will follow in a subsequent PR.
- **Not wired into the layer engine.** Nothing here changes Layer Manager or how `X-Env-Layer-Provides`/`RequiresProvider` behave - no existing layer or build is affected.

Net effect: purely additive, self-contained (`trait/` directory, one new module, plus a handful of extensions to existing shared parsing/validator code). **No behavioural change to any current build.**

### DIFF
--- a/docs/config/index.adoc
+++ b/docs/config/index.adoc
@@ -9,7 +9,7 @@ link:../index.adoc[← Back to Main Documentation]
 
 == Overview
 
-rpi-image-gen provides a flexible, hierarchical approach to managing build variables through multiple configuration sources with clear precedence rules. The system is built around the `IGconf_` prefix namespace and supports both INI and YAML file formats.
+rpi-image-gen provides a flexible, hierarchical approach to managing build variables through multiple configuration sources with clear precedence rules. The system is built around the `IGconf_` prefix namespace and uses YAML configuration files.
 
 == Core Architecture
 
@@ -131,13 +131,9 @@ FINAL ENV
 
 == Configuration File Support
 
-The configuration system supports both INI and YAML file formats with automatic format detection based on file extension.
+The configuration system uses YAML configuration files with native support for hierarchical data and includes.
 
-=== YAML Configuration Format
-
-YAML provides the most flexible configuration format with native support for hierarchical data and includes.
-
-==== Mapping Sections
+=== Mapping Sections
 
 [source,yaml]
 ----
@@ -158,7 +154,7 @@ image:
   root_part_size: 300%
 ----
 
-==== List Sections
+=== List Sections
 
 In addition to mapping (key-value) sections, YAML configuration files support sequence (list) sections. List items are assigned 0-based numeric keys, producing the same `IGconf_<section>_<key>` variable format as mapping sections.
 
@@ -198,31 +194,9 @@ List sections do not merge with includes in the same way as mapping sections. Be
 ====
 
 
-=== INI Configuration Format
-
-INI format provides traditional section-based configuration:
-
-[source,ini]
-----
-!include base.cfg
-
-[env]
-MYVAR = UNCHANGED
-
-[device]
-class = pi5
-storage_type = sd
-hostname = ${IGconf_device_class}-$(tr -dc 'a-z' < /dev/urandom | head -c 6)
-
-[image]
-compression = zstd
-boot_part_size = 200%
-root_part_size = 300%
-----
-
 === Configuration Includes
 
-Both formats support including other configuration files to promote reusability and modularity. The include precedence order is shown in the table below.
+Configuration files support includes to promote reusability and modularity. The include precedence order is shown in the table below.
 
 [cols="1,6,5"]
 |===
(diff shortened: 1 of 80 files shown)


learmj:
Hi @pelwell @popcornmix I'd appreciate if you could review the registry files for factual accuracy and correctness, and have a play with the `config --trait` CLI command to expand a trait.

Once hooked up to the layer engine, a single declaration of a trait like `hw:device:rpi:cm5` will populate the provider index with all expanded tokens, therefore giving visibility of all those to the build.

pelwell:
The querying from the command line works really well - it makes the whole system much more user-friendly.

There is a danger of going too far trying to describe everything, but I think I've found a small hole. Before reading the description I was thinking of `hw:pcie` as describing the hardware interface, so I was surprised not to see it activated by the bcm2712; now I understand that it is about the connector - in which case, is it reasonable that `hw:storage:nvme` requires it?

I think `hw:pcie` should be a trait of the SoC - bcm2712 and bcm2711. We then need ways to say that:
* NVMe requires an M.2 socket.
* Pi 500 has an M.2 socket, as does the CM5IO board.
* The M.2 HAT+ provides an M.2 socket for Pi 5.
etc.

learmj:
Thanks. PCIe did make me scratch my head more than anything else in the current registry.
I don't think `hw:pcie` should move to the soc level _with it's current description_. Is there a case where "this soc has PCIe capable silicon" matters independently of "there's an accessible PCIe connector"? Plus, any board with an M.2 socket is on PCIe capable hardware anyway (soc / bridge chip, etc).

In the case of a custom / OEM CM5 carrier board, it's conceivable that it wouldn't have a physical PCIe interface brought out. In that case, although it's onboard the module and routed out via the connectors, it wouldn't feel right having to disable it to indicate the physical interface isn't there (because it would be enabled if declared by `hw:soc:bcm2712`).

> NVMe requires an M.2 socket.

NVMe relies on PCIe but it doesn't always require an M.2 socket - M.2 is just the connector. 
`hw:storage:nvme` requiring `hw:pcie` as dependency is the correct minimal dependency.

If we were to move `hw:pcie` into the soc (which I'm not against _if_ the description is changed), I think we should then add:

`hw:pcie:m2` - PCIe capable M.2 socket available on this board
`hw:pcie:socket` - A physical PCIe interface connector exists on this board

The description for `hw:pcie` could be changed to, "PCIe bus presence" or similar.

`hw:pcie:socket` would have a dependency on `hw:pcie`.

In the trait def for `hw:pcie:m2` we could have:
`Triggers: when=y set (hw:pcie:socket)=y`

For CM5IO, we would enable `hw:pcie:m2`.
Same for Pi5, if the configuration uses an  M.2HAT. Pi 500 could be described this way too.

Currently, there aren't traits for Compute Module carrier boards. I have local commits which break out the config files into fragments, eg cm5.yaml, cm5-lite.yaml, cm5io.yaml etc. My current cm5io.yaml just has:
```
(feat/traits)$ cat ./config/device/cm5io.yaml 
# Specifically describes the official Raspberry Pi CM5 IO carrier board.
# A custom/OEM carrier almost certainly differs - ethernet, PCIe routing
# etc are properties of the carrier, not the module itself (see trait.adoc).
# Use this file as a template, not a drop-in: declare your own carrier's
# actual traits in your own config.
trait:
  hw:ethernet: true
  hw:pcie: true
```
As there are so many different carrier boards (ours + OEM) I though it best for them to be described simply like the above, rather than by perhaps enforcing their use via traits. Ultimately, it's the interface capability description that matters.
I'm not adverse to having specific traits for our carrier boards if genuinly needed.

I'm really reluctant to go too far right now when trying to describe things via traits in lots of detail, particularly as our sw and hw is very 'plug n play' capable. Conversely, it's important that the traits are able to clearly express different hw ocnfigurations.

learmj:
Hi @pelwell This follow-on commit should (I hope) address all the above. Let me know what you think.

pelwell:
I was expecting `bcm2711` and `bcm2712` to activate `hw:pcie:bus`.

learmj:
It's activated automatically by parent inference via the device trait. From a build sys perspective, do we need to know that a soc is pcie capable? Not sure what we gain?

learmj:
We could add a hw:soc:pcie if you think it's worth having a SoC specific PCIe token?

pelwell (approved):

PR #283, opened 2026-07-16
examples: fix merged-usr firmware paths

PaulMarisOUMary:
## Summary

`examples/slim/device/mypi5/myboard.yaml` uses `dpkgopts` rules to trim the firmware installed by `firmware-brcm80211` down to the single chip used by Pi 5.
On Debian Trixie, these rules have no effect at all: the untrimmed firmware set for every Raspberry Pi model (plus several unrelated other boards) silently ends up in the built image, with no error at build time.

## Root cause

https://github.com/raspberrypi/rpi-image-gen/blob/510ff06b1514a8dacea7426da52ae9ff7df7ae29/examples/slim/device/mypi5/myboard.yaml#L36-L41

These rules target `/lib/firmware/...`, which matches the literal path recorded in the `firmware-brcm80211` `.deb` itself:

```
$ dpkg -c firmware-brcm80211_*.deb | grep raspberry
lrwxrwxrwx root/root 0 ... ./lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,5-model-b.bin -> ../cypress/cyfmac43455-sdio.bin
...
```

However, Debian's merged-usr transition means `dpkg` now canonicalizes `/lib/*` to `/usr/lib/*` before evaluating `--path-exclude` & `--path-include` glob patterns.
A rule written against `/lib/...` therefore matches nothing.

## How to reproduce

On `master`, build `examples/slim/config/pi5-slim.yaml`, then:
```
$ ls -1 /usr/lib/firmware/brcm/*raspberry*.bin
/usr/lib/firmware/brcm/brcmfmac43430-sdio.raspberrypi,0-compute-module.bin
/usr/lib/firmware/brcm/brcmfmac43430-sdio.raspberrypi,3-model-b.bin
/usr/lib/firmware/brcm/brcmfmac43430-sdio.raspberrypi,model-zero-2-w.bin
/usr/lib/firmware/brcm/brcmfmac43430-sdio.raspberrypi,model-zero-w.bin
/usr/lib/firmware/brcm/brcmfmac43430b0-sdio.raspberrypi,model-zero-2-w.bin
/usr/lib/firmware/brcm/brcmfmac43436-sdio.raspberrypi,model-zero-2-w.bin
/usr/lib/firmware/brcm/brcmfmac43436s-sdio.raspberrypi,0-compute-module.bin
/usr/lib/firmware/brcm/brcmfmac43436s-sdio.raspberrypi,model-zero-2-w.bin
/usr/lib/firmware/brcm/brcmfmac43439-sdio.raspberrypi,0-compute-module.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,3-model-a-plus.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,3-model-b-plus.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,4-compute-module.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,4-model-b.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,5-compute-module.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,5-model-b.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,500.bin
/usr/lib/firmware/brcm/brcmfmac43456-sdio.raspberrypi,4-compute-module.bin
/usr/lib/firmware/brcm/brcmfmac43456-sdio.raspberrypi,400.bin
```

Every Pi model is present, not just the Pi 5.

## Fix

Rewritting the rules from `/lib/firmware` to `/usr/lib/firmware` resolves it.
Rebuilding the example now produces the expected output:

```
$ ls -1 /usr/lib/firmware/brcm/*raspberry*.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,5-compute-module.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,5-model-b.bin
/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,500.bin
```

## Note

Would it make more sense to fix this at the tool level instead?
A helper that resolves `path-exclude` & `path-include` rules against what's actually recorded in the `.deb`, rather than the eventual on-disk target path, would avoid this whole class of merged-usr surprise, here and in any future layer.
Happy to look into that if it's a direction you'd want, either way, this PR fixes the immediate issue.

### DIFF
--- a/examples/slim/device/mypi5/myboard.yaml
+++ b/examples/slim/device/mypi5/myboard.yaml
@@ -33,12 +33,12 @@ mmdebstrap:
     - path-exclude=/usr/lib/linux-image-*-v8/broadcom/bcm2837*
     - path-exclude=/usr/share/doc/firmware-brcm80211/
     - path-include=/usr/share/doc/firmware-brcm80211/copyright
-    - path-exclude=/lib/firmware/cypress/*
-    - path-include=/lib/firmware/cypress/cyfmac43455*
-    - path-exclude=/lib/firmware/brcm/*
-    - path-include=/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,5*
-    - path-include=/lib/firmware/brcm/brcmfmac43455-sdio.txt
-    - path-include=/lib/firmware/brcm/BCM4345C0.raspberrypi,5*
+    - path-exclude=/usr/lib/firmware/cypress/*
+    - path-include=/usr/lib/firmware/cypress/cyfmac43455*
+    - path-exclude=/usr/lib/firmware/brcm/*
+    - path-include=/usr/lib/firmware/brcm/brcmfmac43455-sdio.raspberrypi,5*
+    - path-include=/usr/lib/firmware/brcm/brcmfmac43455-sdio.txt
+    - path-include=/usr/lib/firmware/brcm/BCM4345C0.raspberrypi,5*
   customize-hooks:
     - mkdir -p $1/boot/firmware
     - ln -sf firmware/overlays $1/boot/overlays


learmj:
@PaulMarisOUMary

> A helper that resolves path-exclude & path-include rules against what's actually recorded in the .deb, rather than the eventual on-disk target path, would avoid this whole class of merged-usr surprise, here and in any future layer. Happy to look into that if it's a direction you'd want

How would you actually implement that? I'd like to hear your thoughts. You'd have to hook into mmdebstrap's processing of each deb. As mmdebstrap owns the entire apt orchestration, I really don't that's possible. Sounds more of an AI generated offering to me...

Nevertheless, it's a genuine trivial issue you've spotted with the dpkgopts and as the example specifically targets trixie, changing the filtering to target `/usr/lib/firmware/xxx` instead of `/lib/firmware/xxx` is fine.

PaulMarisOUMary:
@learmj 

That's fair on mmdebstrap. 
Here's how I actually got to that note though, wasn't pulled out of nowhere. The fix itself is simple, `/lib` -> `/usr/lib`, but that alone won't stop the next person hitting the same thing in their own layer. I also considered just auto-rewriting any `/lib` path in `dpkgopts` to `/usr/lib`, but that felt like dodging the actual issue.

So I looked at how this rule ended up wrong in the first place. The `.deb` (`dpkg -c firmware-brcm80211*.deb`), literally lists `/lib/firmware/...`, so my guess is whoever wrote this just mirrored the package content. Which led me to wonder if excludes/includes could be derived straight from the `.deb` instead of being set post-install.

Turns out that's not reliable either. What's listed in a `.deb` isn't what ends up on disk. `.postinst` scripts, merge-usr and similar, can all move things between "declared in the package" and "actually there after install".

Only real way to know for sure is `dpkg -L firmware-brcm80211` on an actual build. Not sure there's a practical way to check this without one though, curious if you've got a better angle.

From what I explored, there's no simple way to build exclude/include resolution off package content. Building tooling around seems like a lot for not much, and since rpi-image-gen only targets Debian anyway, documenting it seems like the actual fix. Probably belongs in `LAYER_BEST_PRACTICES` or somewhere in `/docs`, would've saved me this whole detour.

PR #296, opened 2026-08-04
device: Inject traits and add config fragments

learmj:
Initial introduction of trait tokens across various layers, primarily device related.
Config fragments added to illustrate using traits to describe platforms/features although not consumed by any top level config files yet.


learmj:
Hi @pelwell Does this look reasonable?

pelwell (approved):
I like it - it's very readable, and it seems expressive enough.

PR #301, opened 2026-08-10
image-rota: use bootfs mountpoint for BOOTCONFIG partition

mricher-git:
Uses /bootfs mountpoint instead of hardcoded file autoboot.txt to allow other layers to seed files in the boot partition

### DIFF
--- a/image/gpt/ab_userdata/image.yaml
+++ b/image/gpt/ab_userdata/image.yaml
@@ -3,7 +3,7 @@
 # X-Env-Layer-Category: image
 # X-Env-Layer-Desc: Immutable GPT A/B layout for rotational OTA updates,
 #  boot/system redundancy, and a shared persistent data partition.
-# X-Env-Layer-Version: 5.5.1
+# X-Env-Layer-Version: 5.6.0
 # X-Env-Layer-Requires: image-base,device-base,rpi-ab-slot-mapper,systemd-min
 # X-Env-Layer-Provides: image
 #
@@ -81,3 +81,5 @@ mmdebstrap:
     - e2fsprogs
     - initramfs-tools
     - rsync
+  customize-hooks:
+    - mkdir -p $1/bootfs
--- a/image/gpt/ab_userdata/pre-image.sh
+++ b/image/gpt/ab_userdata/pre-image.sh
@@ -30,7 +30,7 @@ VERITY_ARGS_SYSTEM="\
 # Set up the partition layout for tryboot support. Partition numbering
 # relates directly to the layout in genimage.cfg.in
 
-cat << EOF > "${genimg_in}/autoboot.txt"
+cat << EOF > "${fs}/bootfs/autoboot.txt"
 [all]
 tryboot_a_b=1
 boot_partition=2
--- a/image/gpt/ab_userdata/genimage.cfg.in.erofs
+++ b/image/gpt/ab_userdata/genimage.cfg.in.erofs
@@ -78,10 +78,10 @@ image <IMAGE_NAME>.<IMAGE_SUFFIX> {
 image bootconfig.vfat {
    vfat {
       label = "BOOTCONFIG"
-      file "autoboot.txt" { image = "autoboot.txt" }
       extraargs = "-s 1 -S <SECTOR_SIZE>"
    }
    size = 32M
+   mountpoint = "/bootfs"
 }
 
 image boot.vfat {
--- a/image/gpt/ab_userdata/genimage.cfg.in.ext4
+++ b/image/gpt/ab_userdata/genimage.cfg.in.ext4
@@ -78,10 +78,10 @@ image <IMAGE_NAME>.<IMAGE_SUFFIX> {
 image bootconfig.vfat {
    vfat {
       label = "BOOTCONFIG"
-      file "autoboot.txt" { image = "autoboot.txt" }
       extraargs = "-s 1 -S <SECTOR_SIZE>"
    }
    size = 32M
+   mountpoint = "/bootfs"
 }
 
 image boot.vfat {


learmj:
Thanks @mricher-git As I mentioned elsewhere, I agree with this in principle. I tried something similar recently, but I wasn't able to pack a vfat image from a particular non-root mount point without a genmage patch to enabled size in percent to be used. See https://github.com/pengutronix/genimage/pull/340 for further details. In that case, the image I was creating had no need for a fixed size - on the contrary - it was important that it was allowed to vary, but needed no headroom.

We need fixed size partitions for AB, so your change will work as-is since we are pinning the size to an absolute value and current genimage is happy with that. I'm unsure if I'd prefer to wait until upstream genimage merges the proposed PR I linked to before taking yours. It probably doesn't make sense to wait.

One item I'm currently working on is a rescue image base layer which creates a tiny bootable vfat filesystem and initramfs from `/bootfs`, but I can't set that image to a fixed size as users must be able to add things to the boot disk contents (packed from `/bootfs`), so the size must be `100%`.

mricher-git:
Albeit a simple 5 line edit, hopefully now you can see how this spiraled into my Design Philosophy post, with hindsight 20/20 had I just made my edits in a vendored clone vs the franken-layer I composed, a simply `git pull` would have easily picked up your fix. I should have just PRed this from the  beginning, hopefully each of these build knowledge on a solid layer override framework.

Personally, I'm good for now if you feel the right thing to do is wait.

image/gpt/ab_userdata/pre-image.sh
```diff
@@ -30,7 +30,9 @@ VERITY_ARGS_SYSTEM="\
 # Set up the partition layout for tryboot support. Partition numbering
 # relates directly to the layout in genimage.cfg.in
 
-cat << EOF > "${genimg_in}/autoboot.txt"
+mkdir -p ${fs}/bootfs
```
learmj: A minor nit is that because `/bootfs` is created in `pre-image`, a layer author who inherits `image-rota` and who wants to install files into `/bootfs` so they are present in the boot filesystem, would need to mkdir `$1/bootfs` in their own layer before copying in files. This goes against the composable nature of things somewhat, although I appreciate it's often hard to know what to do for the best to keep things as 'composable as possible'. Suggest using a customize hook in the layer yaml instead to do the mkdir.

image/gpt/ab_userdata/image.yaml
```diff
@@ -3,7 +3,7 @@
 # X-Env-Layer-Category: image
 # X-Env-Layer-Desc: Immutable GPT A/B layout for rotational OTA updates,
 #  boot/system redundancy, and a shared persistent data partition.
-# X-Env-Layer-Version: 5.5.1
+# X-Env-Layer-Version: 5.6.0
```
learmj, line 6: Suggest adding this instead of the pre-image mkdir:
```
+  customize-hooks:
+    - mkdir -p $1/bootfs
```

learmj:
If you think the suggested changes are ok and are happy to push an updated PR, I'll merge it.

image/gpt/ab_userdata/pre-image.sh
mricher-git: Does rhe rootfs-overlay/bootfs not handle that case?

I thought I saw a recent changelog about overlays coming before hooks so hooks could operate on overlaid files.

Perhaps we go with the hook either way and nuke the .empty because the .empty currently ends up in the final image

learmj: I'm planning to restructure the discrete hooks soon. There will be breaking changes unless I can easily and sensibly retain backward compat for a while.
A sync of an overlay can be updated to skip files like .gitkeep so yes, an overlay is another scheme that can be used to seed that directory in a way that still adheres to a composable model. Personally, hooks invoked by bin/runner seem a bit 'hidden' and under the covers, but sometimes needs must.

mricher-git: Apologies, it's not clear to me what the final decision is. It may not be obvious since it's not part of my commit; the `rootfs-overlay/bootfs` already exists (and contains a `.empty`)

Is your preference to:
1. keep that as is
2. remove it and add the hook instead
3. keep it and also add the hook

I assume remove mkdir -p from pre-image.sh in all cases.

learmj: I'm sorry for not being clear. Option 2 is my preference. That way, it's clear to anyone reading the YAML that `/bootfs` has been created and if inheriting / depending on this layer, it can be filled with content that will reside in the bootfs image itself. Since https://github.com/raspberrypi/rpi-image-gen/commit/9ae86e1485a10496de4f6c463169e233a73e0054, overlays are applied _before_ a layer's `customize` hooks run, so having `/bootfs` in the overlay works fine too. There is no functional difference. I personally view the layer YAML as 'top level' readable material whereas hooks and overlays are somewhat obfuscated.

learmj (approved):

learmj:
@mricher-git You can also safely remove `image/gpt/ab_userdata/device/rootfs-overlay/bootfs/`

mricher-git:
Screenshot_20260812_170839_Chrome

@learmj  I removed the .empty, that should be sufficient to nuke the rootfs-overlay/bootfs

learmj (approved):
LGTM. Thank you very much.
I'm sure I grabbed the raw diff and didn't see the `.empty` as being deleted, but I've repeated that again and it is truly gone. GitHub just renders it as "Whitespace-only changes."

mricher-git:
For sure the UI doesn't make it obvious, I struggled to prove it was gone 😄 Thanks for pulling this so quickly, helps a lot.

PR #302, opened 2026-08-11
layer: share network-manager persistent state across slots

### LINKED ISSUE #300: NetworkManager - /var/lib/NetworkManager state w/image-rota should have shared slot
I believe /var/lib/NetworkManager should have a shared slot config for /var/lib/NetworkManager. It contains secret_key, lease info etc, and the files in there are more to do with the persistent state of the device, not per slot or runtime info. I believe nm docs state both /etc/machine-id and /var/lib/NetworkManager should be backed-up per machine for consistency with DHCP leases (both ways) which would further strengthen the idea.

learmj:
/var/lib/NetworkManager holds secret_key and lease/state data that need to stay consistent across a slot flip, eg stable privacy IPv6, DHCP lease continuity, etc. The same continuity requirement is already covered for NM system-connections. NM's slot-shared config now contains two paths making both retained across slot rotations.

Document why /var defaults to slot-local (ie per-slot) rather than shared for future layer authors deciding whether they need a slot-shared.d entry.

Resolves #300

### DIFF
--- a/docs/layer/image-rota.html
+++ b/docs/layer/image-rota.html
@@ -302,6 +302,21 @@
 <p>Any layer can declare filesystem paths whose application data should be shared across slots. At boot, each declared path is bind-mounted from a common location on the persistent partition so the same data is visible regardless of which slot is active.</p>
 </div>
 <div class="sect2">
+<h3 id="_why_var_is_slot_local_by_default">Why /var is slot-local by default</h3>
+<div class="paragraph">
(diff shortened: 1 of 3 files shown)


learmj:
@mricher-git do you approve?

mricher-git:
I think the documentation does a great job at explaining and defending the position, well done!

NM tweak looks good.

Issue #212, opened 2026-03-23
Feature request: add a wifi layer (PR contribution offer)


johanehnberg:
While #207 would bring a simple wifi station example configuration, it is limited to station only. After the work on that documentation work is finished, I am interested in contributing a more complete wifi configuration layer.

Currently, I have working script for managing wifi using the official approach from https://www.raspberrypi.com/documentation/computers/configuration.html. Three modes are working currently: station (client), AP router, and AP bridge. It has built-in support for RPi-specific optimizations and manages some upstream bugs. Turning it into a layer should be quite doable if desired. It seems to have at least some interest with #47 and #64.

learmj:
Hi @johanehnberg Sounds reasonable, but I think it would be challenging to provide generic solutions for this.
As part of #184 I plan on creating some new layers with specific [providers](https://raspberrypi.github.io/rpi-image-gen/layer/index.html#_x_env_layer_provides_x_env_layer_requiresprovider) and have these use some cloud init terminology, eg, such as '_network-renderer_' (something that translates network configuration files to files understood by a network mgmt stack), '_network-activator_' (something that controls network interfaces based on a rendered configuration). This is probably orthogonal to actual configuration layers.
Perhaps contributions into `contrib/` would be more applicable for what you are describing.

johanehnberg:
Sure, that is considerably lower effort. It will then be a contribution specific to Trixie + network-manager + Raspberry Pi official recommendation + community best practices + real-world bug workarounds. I will target this issue with the PR.

learmj:
What do you think of the recent wlan specific additions?

johanehnberg:
I have not tested them in practice yet, but I am very happy to see wlan capabilities becoming part of the built-in layers.

There is likely a lot of thought behind the design choices, which is naturally not visible in code. With these few questions, I could provide a bit more insight on the implementation:

- The choice of networkd over networkmanager is interesting - the official docs go for networkmanager, why deviate?
- The choice of iwd is also curious - as above, and are there any known limitations?
- What are the intended use cases, or, is the idea to make it possible to add functionality to anything wlan?

learmj:
rpi-image-gen is a general purpose image builder and one of the aims of the project is to try and not make assumptions about use cases, or impose certain ways of doing things, hard coupling etc. It's the primary tool we offer for embedded / industrial image builders simply because of the customisation it supports, and how it can scale to support different configurations / uses.

[...]

PR #305, opened 2026-08-13
wireless-regulatory: default regdom to DFS-UNSET

learmj:
Code 00 is a safe global default. GB is not.

Refs #212

### DIFF
--- a/layer/net-wireless/regulatory.yaml
+++ b/layer/net-wireless/regulatory.yaml
@@ -2,12 +2,12 @@
 # X-Env-Layer-Name: wireless-regulatory
 # X-Env-Layer-Category: connectivity
 # X-Env-Layer-Desc: Wireless regulatory database
-# X-Env-Layer-Version: 1.1.0
+# X-Env-Layer-Version: 1.2.0
 # X-Env-Layer-Provides: wireless-regdb
 #
 # X-Env-VarPrefix: ieee80211
 #
-# X-Env-Var-regdom: GB
+# X-Env-Var-regdom: 00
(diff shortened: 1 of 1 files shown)


learmj:
fyi @mricher-git @johanehnberg

mricher-git:
@learmj I had this same fix locally and learned I had to wrap 00 in quotes otherwise only a single 0 written

learmj:
> @learmj I had this same fix locally and learned I had to wrap 00 in quotes otherwise only a single 0 written

Umm
```
$ cat /etc/modprobe.d/cfg80211_regdomain.conf
options cfg80211 ieee80211_regdom=00
```

mricher-git:
Oh right, my bad, it was when I was specifying it in my config:
```
ieee80211:
  regdom: '00'
```

learmj:
argh... ConfigLoader._load_yaml reads the config file with yaml.safe_load(). PyYAML uses the YAML 1.1 core schema, whose resolver treats a bare scalar that looks like a leading-zero octal literal as an integer. Basically, this means that an unquoted 00 parses as the Python int 0, not the string "00". This is a YAML parsing quirk not a bug in the config loader (IMHO).

Wonder if I should change `X-Env-Var-regdom-Valid` from from `string` to `regex:^[A-Z0-9]{2}$`. I think that would cover all codes.. Thoughts?

learmj:
`regex: ^(00|[A-Z]{2})$` is probably better (safer).

mricher-git:
My 2c is I think your earlier suggestion is potentially better. It's purely syntactic validation, not current valid regdom validation making it more future proof.

PR #306, opened 2026-08-13
wireless-regulatory: validate regdom with regex

learmj:
The string validator allows bogus country codes like 5B or A1 but also means that a setting of 00 via config file lands as 0 in /etc/modprobe.d/cfg80211_regdomain.conf because PyYAML's resolver treats a bare scalar that looks like a leading-zero octal literal, as an integer. We can better enforce the Regulatory Domain using a regex to stop this happening, and also hard fail on bogus values.

Refs #305
https://wireless.docs.kernel.org/en/latest/en/developers/regulatory.html

### DIFF
--- a/layer/net-wireless/regulatory.yaml
+++ b/layer/net-wireless/regulatory.yaml
@@ -2,7 +2,7 @@
 # X-Env-Layer-Name: wireless-regulatory
 # X-Env-Layer-Category: connectivity
 # X-Env-Layer-Desc: Wireless regulatory database
-# X-Env-Layer-Version: 1.2.0
+# X-Env-Layer-Version: 1.3.0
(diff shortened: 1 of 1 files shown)


learmj:
@mricher-git This should help. At least without quotes, `00` coming in via the config file now fails validation rather than silently passing and `0` landing in the image.
```
Error: [FAIL] IGconf_ieee80211_regdom=0 (invalid value, reason: Value '0' does not match pattern: ^(00|[A-Z]{2})$, layer: wireless-regulatory)
```

learmj:
As mentioned by @mricher-git in #305 we shouldn't enforce the regdom too much (it's not ours to define). Relax the regex to validate that `regdom` is 2 uppercase alphanumeric characters.

Issue #311, opened 2026-08-18
Restructure documentation for better access to layers


mricher-git:
After getting comfortable with how the tool works, the most relevant and frequently accessed docs would likely be `Available Layers` under [Layers](https://raspberrypi.github.io/rpi-image-gen/layer/index.html). As it stands it's currently:
1. Buried at the end of "How Layers Work"
2. The TOC for that page is in fact missing the entry/jump to `Available Layers`
3. It's at the end of the doc, making it a little difficult to find quickly

I would suggest carving that out to it's own layer_reference.html and adding `Layer Reference` section as a first class citizen under `Documentation Structure` on the root index. Quick to access and easy to bookmark.

learmj:
Ack. Thanks @mricher-git

PR #317, opened 2026-09-01
per-layer work dir, doc improvements

learmj:
Add support for a per-layer work directory, available to layer metadata via new placeholder `${WORKDIR}` and to hooks as environment variable `${LAYER_WORKDIR}`. Each layer's work dir is created automatically before bdebstrap execution takes place and is not wiped across builds (delete manually or use the `clean` command to erase).

A work directory directory provides a general scratch pad location for layers to use, eg for build/compilation operations. It's also now where the auto-generated pre/post YAML fragments live (ie, which support per-layer `customize` phase filesystem overlays).

Layers are already keyed by name + version, with a clash resulting in a hard error. This same key enables us to create a unique and clearly named path for each work dir under `IGconf_sys_workroot`. The engine emits a tagged, rather than relative, path for these directories (eg `WORKROOT:build/layer/image-rota-5.6.0`) because at the point the engine finishes execution, `IGconf_sys_workroot` has not been resolved. `lib/common.sh::map_path()` allows the tag to be mapped at run-time when the build commences, and so the full on-disk path is available then.

#311 requested that the layer reference page be standalone - it was becoming unwieldy to have as part of the layer's own docs page and had no quick nav. Both now solved.

### DIFF
--- a/bin/runner
+++ b/bin/runner
@@ -126,7 +126,7 @@ run_dir_hooks() {
 
 # Per-layer phased execution
 run_layer_step() {
-   local layer=$1 version=$2 stempath=$3 phase=$4; shift 4
+   local layer=$1 version=$2 stempath=$3 resolved=$4 workdir=$5 phase=$6; shift 6
    local dest=${1:-}
    local ldir vmajor vminor vpatch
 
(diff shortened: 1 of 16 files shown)


mricher-git:
@learmj 

https://github.com/raspberrypi/rpi-image-gen/blob/master/docs/index.adoc is referencing the layer reference by .html instead of .adoc

404 - page not found
The master branch of rpi-image-gen does not contain the path docs/layer_reference.html

learmj:
It's ok. I've not generated the docs yet.

learmj:
I misread your comment. You're right. I totally missed that as I was only looking at the html :-D

learmj:
@mricher-git see:
https://learmj.github.io/rpi-image-gen/index.html
and https://learmj.github.io/rpi-image-gen/layer_reference.html

This is what will merge to master.

Issue #63, opened 2025-05-19
Add layer for running only from RAM


septatrix:
To improve longevity of the storage medium it is often desirable to run entirely from RAM (as long as the software is small enough and does not consume a lot of memory). In the wild there seem to be a few tools to convert a running image to one which run primarily from RAM but it would be favorable if one could built an image with such a setup from the start

learmj:
I agree it would be nice to have an example to create a small initramfs only based image.

In the interim, if you need this functionality, take a look at our https://github.com/raspberrypi/pi-gen-micro

septatrix:
> In the interim, if you need this functionality, take a look at our https://github.com/raspberrypi/pi-gen-micro

Our image is a bit bigger as it also needs to include a GUI. Is there a limit on the initrd or boot partition size from the point of the RPi bootloader/firmware (for RPi 4)?

tdewey-rpi:
> > In the interim, if you need this functionality, take a look at our https://github.com/raspberrypi/pi-gen-micro
> 
> Our image is a bit bigger as it also needs to include a GUI. Is there a limit on the initrd or boot partition size from the point of the RPi bootloader/firmware (for RPi 4)?

Yes - circa 96MiB for the initramfs.

You can indeed build bigger things with pi-gen-micro, but it's more involved (and with a bigger focus on creating minimal-footprint images). In the future, we may look to introduce examples in rpi-image-gen, but there is a definite intent to migrate our embedded Raspberry Pi Imager to pi-gen-micro, which may serve as a template on a shorter timeframe.

Have you contacted applications@ about your design? They may be able to offer greater advice & assistance.

septatrix:
> > Our image is a bit bigger as it also needs to include a GUI. Is there a limit on the initrd or boot partition size from the point of the RPi bootloader/firmware (for RPi 4)?
> 
> Yes - circa 96MiB for the initramfs.

Okay, I fear that is not enough so will likely need a minimal initramfs which creates a ramfs and loads everything from the normal partition

> You can indeed build bigger things with pi-gen-micro, but it's more involved (and with a bigger focus on creating minimal-footprint images). In the future, we may look to introduce examples in rpi-image-gen, but there is a definite intent to migrate our embedded Raspberry Pi Imager to pi-gen-micro, which may serve as a template on a shorter timeframe.

[...]

PR #326, opened 2026-09-16
Add microboot image support

learmj:
Microboot is a flavour of image that can be created for specific device deployments that require a tiny footprint. It's not intended to supersede similar images provided by existing Raspberry Pi tools but rather to provide an alternative modular/composable base that can be used in different applications where load time, boot time, size and running from RAM are important considerations. For example gadgets, rescue images, repartitioning operations, attestation stages, etc.


learmj:
Related: https://github.com/raspberrypi/rpi-image-gen/issues/63 fyi @septatrix @bfreis @leinelissen
A little late but may be of interest to you.

